Issue #36, opened 2021-03-08
Work with CDT to set up our initial infrastructure in Azure


kalvinwang:
### Description

We need CDT to set up new environments for us in Azure (test, staging, and production, though we only need staging at this point.) Last year we only had staging and production, which made it tricky for us to enable EDD to test over several days in an environment that wouldn't change.

- Deploys to staging should trigger upon merging a commit to the main branch
- Deploys to test should be manual. 
- Deploys to production should be automatic each morning at 9AM ET from the staging environment, but this will likely need to be changed to manual prior to launch.

### Acceptance Criteria

  - [x] Must be able to deploy application to staging
  - [ ]  staging env should be set up by CDT, integrated with this repo, and include the same set-up-by-CDT features as cagov/unemployment. Current environments should be documented in the technical foundations doc, as well as tickets created for other environments not yet created, if necessary

lomky:
App up and running in staging!

Conclusions:
  - Azure required the full path as deployed to the next executable - relative paths don't work
  - we need to run on port 8080

<details><summary>debug worklog</summary>

[Pipelines](https://calenterprise.visualstudio.com/UI-Claim-Tracker/_build?definitionId=596&_a=summary)
[Running logs](https://portal.azure.com/#@digitalca.onmicrosoft.com/resource/subscriptions/005fa005-b981-47d8-8b39-5b82adf6569c/resourceGroups/rg-edd-pub-vip-claimtrackerui-w-p-001/providers/Microsoft.Web/sites/as-edd-pub-vip-claimtrackerui-w-p-001/slots/Staging/logStream)
[SSH (currently inaccessible)](https://portal.azure.com/#@digitalca.onmicrosoft.com/resource/subscriptions/005fa005-b981-47d8-8b39-5b82adf6569c/resourceGroups/rg-edd-pub-vip-claimtrackerui-w-p-001/providers/Microsoft.Web/sites/as-edd-pub-vip-claimtrackerui-w-p-001/slots/Staging/ssh)

[Staging Site]()


spun up. current state:
 - [github](https://github.com/cagov/ui-claim-tracker/commit/cd0cf34db31e61e3c91f6c942d56e0b010ffbee4)

 app not starting, log shows the message:

 ```
/home/LogFiles/2021_03_22_RD0003FFA46FD1_default_docker.log  ()
2021-03-22T20:37:48.403639695Z    export PORT=8080
2021-03-22T20:37:48.403643895Z fi
[...]

PR #66, opened 2021-03-23
Configure start command for Azure deployments

lomky:
Configures our production `yarn start` command to work in the Azure environment. See [most recent pipeline](https://calenterprise.visualstudio.com/UI-Claim-Tracker/_build?definitionId=596&_a=summary) and the now running staging environment.

See [full debug worklog](https://github.com/cagov/ui-claim-tracker/issues/36#issuecomment-804981894) for the nitty gritty details.

===

Bumps #36

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "start": "node_modules/next/dist/.bin/next start"
+    "start": "next start"
   },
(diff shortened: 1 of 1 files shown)


kalvinwang (approved):
Nice job! the debug worklog makes for good reading. I wonder why it's port 8080, that doesn't seem to be the default, and we don't seem to have configured Next.js to use port 8080.

PR #151, opened 2021-05-07
Clean up Header MVP & Storybook

lomky:
Updates the Header, removing the name field as its not MVP. 

Plus a quick cleanup of how to stories display and removing the example stories.

Before | After
--|--
Old header with name| New header without name
Old extraneous nested layout | new cleaner story layout


.storybook/preview.js
```diff
@@ -21,7 +21,7 @@ export const parameters = {
   options: {
     storySort: {
       method: 'alphabetical',
-      order: ['Claim Tracker', 'Component', ['Atom', 'Page Section'], 'Example'],
```
lomky, line 24: Sort based on the actual page

components/Header.tsx
```diff
@@ -25,12 +21,6 @@ export const Header: React.FC<HeaderProps> = () => {
           <Nav.Link target="_blank" rel="noopener noreferrer" href="https://askedd.edd.ca.gov/">
             <span className="text">{t('help')}</span>
           </Nav.Link>
-          <Navbar.Collapse>
-            {/*  TODO what is the URL here? */}
-            <Nav.Link target="_blank" rel="noopener noreferrer" href="https://edd.ca.gov">
-              <span className="text">{t('welcome')}</span>
-            </Nav.Link>
-          </Navbar.Collapse>
```
lomky, line 33: This removes the section dealing with a user name

tests/claimstatus/__snapshots__/index.test.tsx.snap
```diff
@@ -62,26 +62,6 @@ exports[`Exemplar react-test-renderer Snapshot test renders homepage unchanged 1
...
-            rel="noopener noreferrer"
-            target="_blank"
-          >
-            <span
-              className="text"
-            >
-              Welcome, name
-            </span>
-          </a>
-        </div>
```
lomky, line 84: Updates the snapshot to remove the header name piece

stories/Breadcrumbs.stories.tsx
```diff
@@ -1,15 +1,15 @@
 import { Story, Meta } from '@storybook/react'
```
lomky, line 1: all `stories/*.stories.tsx` were changed like so:
 - import the component `as <Name>Component`
 - set the story to be `<Name>`

When the only export from a story is the same name as the exported title, it hoists up the display.

kevmo (approved):
🥇 Looks good! 🚢

PR #205, opened 2021-06-02
Adds Shimmer Loading Effect

lomky:
### Description

  - Adds a shimmer component & styling for it
  - Adds shimmerable components for:
    - Key&value fields
    - Text lines
    - Next Steps block
  - Embeds those components in the Claim Section where appropriate
  - Passes loading state down through the component stack from where we make the API call
    - **nb:** its possible there's a more elegant way to do this, but I'm calling Good Enough for now with our looming deadlines
  - Adds stories for each

### Changes
Component | Normal | Loading
-- | -- | --
Main | main app normal | main app loading
Claim Details | image | image
Claim Status | image | image
NextSteps | image | image
InfoField | image | image
TextLine | image | image
Shimmer | - | image
ShimmerField | - | image

### Design Review
The Storybook for Design's review can be accessed [here.](https://60705d04dcad7600211e34d2-kfvehegept.chromatic.com)
See the Page Section Main component, and its Loading state.

### Testing
(tests should pass after #219 merges)
The loading state cannot be invoked naturally the app until we split the page load into two. For now

  - `git fetch && git checkout kat/121_shimmer_loading && git pull`
  - `yarn dev`
  - Navigate to http://localhost:3000/claimstatus - App should load without issue.
  - edit `pages/index.tsx` line 37 to  `loading={true}`
  - Navigate to http://localhost:3000/claimstatus - App should load with the shimmers.

You can also examine the various componenets & page sections listed above in Storybook and toggle Loading to see each one.
  - `git fetch && git checkout kat/121_shimmer_loading && git pull`
  - `yarn storybook`
  
Resolves #121 

- [ ] Claim Status shows a shimmer for pending data
- [ ] Ticket filed for splitting page load

Issue #121, opened 2021-04-23
Create Loading state for Claim Tracker


nguyen-tina:
### Task
We need to define a loading state for the page, in case the internet connection is slow.

### Details / Context
The loading state should work on mobile and desktop, no matter the browser or device. It should be the first thing to load on a page.

### Acceptance Criteria

- [x] Style designed for desktop and mobile

nguyen-tina:
So far, I'm thinking of Figma's loading state-- I like how the bar moves across the page to show progress. Since the global header has a bar at the bottom, I wonder if we can borrow from that experience for our own loading style.

![Screen Shot 2021-04-28 at 1 51 13 PM](https://user-images.githubusercontent.com/17328158/116450019-cc9e0b80-a828-11eb-8ccd-1d2844203b72.png)

I'll reach out to EDD to see what loading styles they have, to sync up for consistency.

nguyen-tina:
There will be three sections of loading time:
- User makes request to UIO + authentication, passes to Azure
- Azure generates page components (very fast)
- Response time of API gateway (actual fields, hopefully should also be fast).

Loading state can be displayed for both second and third sections.

nguyen-tina:
Comments continued in [Slack Thread](https://nava.slack.com/archives/C01Q5640YFR/p1619632351017100)

lomky:
We'll likely be able to use a [drop-in React effect!](https://github.com/gokcan/react-shimmer) for the shimmer sections!

nicoleslaw:
Should we close this out?

lomky:
@nicoleslaw design is done - this is in eng's backlog now.  you can remove the design label maybe?

PR #240, opened 2021-06-15
Adds return type to ShimmerField

lomky:
Cleans up a build warning from GH Actions. See https://github.com/cagov/ui-claim-tracker/pull/205#issuecomment-861020253

Amends #205 #121


kalvinwang (approved):
lgtm, it is strange that the warning only shows up on some PRs...

rocketnova (approved):

PR #244, opened 2021-06-16
Veracode Configuration

lomky:
The veracode github action configuration work in progress.


rocketnova:
@lomky Kalvin and I confirmed today that we are reverting to the original process of having Veracode run as part of the Azure Pipeline, so we can close this PR!

PR #251, opened 2021-06-21
Remove blank lines from azure-pipelines.yml

kalvinwang:
This was originally a test PR to demo something but we do actually want to remove these blank lines!


rocketnova (approved):
🚢

PR #233, opened 2021-06-10
Update Components to SP7

lomky:
Updates to match the site design as of SP7, based on the presentation view on Figma.

Mode | Before | After
--|--|--
Desktop | image | image
Mobile | image | image

Resolves #197 

  - [x] Change Feedback component to an above-header alert banner
  - [x] Remove Breadcrumbs
  - [x] Remove `Manage Claim` Button
  - [x] Proper Heading cadence
    - H1 `Claimtracker`
    - H2 `Claim Status` & `Claim Details`
    - H3 `Next Steps` 
  - [x] Proper vertical spacing between elements
  - [x] Paired with design on final 10% styling efforts
  - [ ] Changes are tested on Staging post-merge into `main`

PR #255, opened 2021-06-22
Add source sans pro to storybook

rocketnova:
## Changes

- Adds `.storybook/preview-head.html`

## Context

We added Source Sans Pro in #233, but it turns out that Storybook doesn't know about the nextjs `<Head>` component. To fix the fonts in Storybook, we need to also [add the font to `<head>`](https://storybook.js.org/docs/ember/configure/story-rendering#adding-to-head) in Storybook.

## Testing Instructions

1. Pull this branch
2. Run `yarn storybook`
3. Verify that Source Sans Pro is being loaded in the Console > Network tab
4. Verify that the text in the iframe is rendered as Source Sans Pro

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,1 @@
+<link href="https://fonts.googleapis.com/css?family=Source Sans Pro" rel="stylesheet" />


lomky (approved):
✨

Issue #242, opened 2021-06-15
Implement business logic


rocketnova:
### Description

Implement business logic to display the correct content for a given claimant, and add the content (untranslated)

### Acceptance Criteria

- [ ] Business logic for scenarios implemented

rocketnova:
Removing the blocked label. I've split this ticket into two:
1. This ticket: Business logic for the scenarios
2. #252: Conditional logic for the pending determination scenario

rocketnova:
The definition for this ticket got too amorphous. I'm adapting this ticket into encompassing:
- #256 
- #263 
- #266 

I'm breaking out new tickets for:
- #290
- #291

PR #256, opened 2021-06-23
(1/?) Implement scenario logic

rocketnova:
## Ticket

#242 

## Changes

- Started adding business logic to identify the correct scenario
- Started adding logic to control content for Claim Status
- Started adding logic to handle passing through Claim Details content
- Created a separate `claim-status.json` namespace for strings related to the Claim Status section
- Created `types/common.tsx` for shared Typescript types
- Added `claim-details.title` to `common.json` for translation consistency
- Removed `extensionEndDate` per conversation with Diana and @nicoleslaw 
- Moved default/placeholder values out of components
- Added test values for components into stories and tests

## Context

This PR begins to the work for #242. There's still quite a bit to do (such as the logic for extracting the Extension Type out of the Program Type, the logic for providing the content for Next Steps, etc), but this PR was already getting too large, so I decided to split the work into multiple PRs.

The work in this PR can be organized into the following broad categories:
- Business logic (see `getScenarioContent.tsx`)
- Adding props to get the content from the business logic into the components
- Breaking out a separate language namespace for the various scenarios that will be displayed in the Claim Status component

## Testing Instructions

- `yarn test`
- `yarn storybook`

### DIFF
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -1,16 +1,10 @@
 import { useTranslation } from 'next-i18next'
 import { InfoField } from './InfoField'
+import { ClaimDetailsContent } from '../types/common'
 
-export interface ClaimDetailsProps {
+export interface ClaimDetailsProps extends ClaimDetailsContent {
   loading: boolean
   title: string
-  programType: string
-  benefitYear: string
-  claimBalance: string
-  weeklyBenefitAmount: string
-  lastPaymentIssued: string
-  extentionType: string
-  extensionEndDate: string
 }
 
 export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -8,13 +8,13 @@ export interface ClaimDetailsProps extends ClaimDetailsContent {
 
 export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
   loading = false,
-  programType = 'Unemployment Insurance (UI)',
-  benefitYear = '3/21/2020 - 3/20/2021',
-  claimBalance = '$508.00',
-  weeklyBenefitAmount = '$120.00',
-  lastPaymentIssued = '4/29/2021',
-  extentionType = 'Tier 2 Extension',
-  extensionEndDate = '5/22/2021',
+  programType,
+  benefitYear,
+  claimBalance,
+  weeklyBenefitAmount,
+  lastPaymentIssued,
+  extensionType,
+  extensionEndDate,
 }) => {
   const { t } = useTranslation('common')
 
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -37,7 +37,7 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
           </div>
           <div className="col-6">
             <div />
-            <InfoField loading={loading} label={t('claim-details.extension-type')} text={extentionType} />
+            <InfoField loading={loading} label={t('claim-details.extension-type')} text={extensionType} />
 
             <InfoField loading={loading} label={t('claim-details.extension-end-date')} text={extensionEndDate} />
           </div>
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -14,7 +14,6 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
   weeklyBenefitAmount,
   lastPaymentIssued,
   extensionType,
-  extensionEndDate,
 }) => {
   const { t } = useTranslation('common')
 
@@ -38,8 +37,6 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
           <div className="col-6">
             <div />
             <InfoField loading={loading} label={t('claim-details.extension-type')} text={extensionType} />
-
-            <InfoField loading={loading} label={t('claim-details.extension-end-date')} text={extensionEndDate} />
           </div>
         </div>
       </div>
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -4,12 +4,10 @@ import { ClaimDetailsContent } from '../types/common'
 
 export interface ClaimDetailsProps extends ClaimDetailsContent {
   loading: boolean
-  title: string
 }
 
 export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
   loading = false,
-  title = 'Claim Details',
   programType = 'Unemployment Insurance (UI)',
   benefitYear = '3/21/2020 - 3/20/2021',
   claimBalance = '$508.00',
@@ -22,7 +20,7 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
 
   return (
     <div className="claim-details container">
-      <h2>{title}</h2>
+      <h2>{t('claim-details.title')}</h2>
 
       <div className="claim-details-box">
         <div className="row">
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -12,9 +12,11 @@ import { Footer } from '../components/Footer'
 import { WorkInProgress } from '../components/WorkInProgress'
 
 import queryApiGateway, { Claim } from '../utils/queryApiGateway'
+import getScenario, { ScenarioType } from '../utils/getScenario'
 
 export interface HomeProps {
   claimData?: Claim[]
+  scenarioType?: ScenarioType
   loading: boolean
 }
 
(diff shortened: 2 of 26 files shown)


components/ClaimDetails.tsx
```diff
@@ -1,34 +1,25 @@
 import { useTranslation } from 'next-i18next'
 import { InfoField } from './InfoField'
+import { ClaimDetailsContent } from '../types/common'
 
-export interface ClaimDetailsProps {
+export interface ClaimDetailsProps extends ClaimDetailsContent {
```
rocketnova, line 5: Note: I used [`extends`](https://www.typescriptlang.org/docs/handbook/2/objects.html#extending-types) on a number of component interfaces so that I could have an interface that was content-only (e.g. `ClaimDetailsContent`) and then have an interface that contained the content as well as the additional props, such as `loading (e.g. `ClaimDetailsProps`). 

Alternatively, if this is hard to read, we can also use nested props, like this:

```tyescript
export interface ClaimDetailsProps {
  loading: boolean
  content: ClaimDetailsContent
}
```
but I thought it was less noisy to have the props on the same level

components/ClaimDetails.tsx
```diff
@@ -1,34 +1,25 @@
...
+  weeklyBenefitAmount,
+  lastPaymentIssued,
+  extensionType,
 }) => {
   const { t } = useTranslation('common')
 
   return (
     <div className="claim-details container">
-      <h2>{title}</h2>
+      <h2>{t('claim-details.title')}</h2>
```
rocketnova, line 22: Note: I changed this from an argument to a string directly referencing the language file because I don't think that the title will conditionally change.

components/ClaimDetails.tsx
```diff
@@ -45,9 +36,7 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
           </div>
           <div className="col-6">
             <div />
-            <InfoField loading={loading} label={t('claim-details.extension-type')} text={extentionType} />
-
-            <InfoField loading={loading} label={t('claim-details.extension-end-date')} text={extensionEndDate} />
```
rocketnova, line 50: Note: I removed this per our conversation today with Diana and @nicoleslaw .

public/locales/en/claim-status.json
```diff
@@ -0,0 +1,11 @@
...
+  "pending-determination": {
+    "description": "We need to confirm your eligibility for benefits."
+  },
+  "base-pending": {
+    "description": "Your payments are pending further review."
+  },
+  "base-no-pending": {
+    "description": "There are no pending issues at this time."
+  }
+}
```
rocketnova, line 11: @nicoleslaw: As a first pass at this, I'm trying out breaking out all the Claim Status scenarios into a `claim-status.json` file. Within the file, each scenario is a top level object. Let me know what you think about this approach! I'm very open to changing it.

stories/ClaimSection.stories.tsx
```diff
@@ -13,6 +14,6 @@ const Template: Story<ClaimSectionProps> = (args) => <ClaimSectionComponent {...
 
 export const ClaimSection = Template.bind({})
 ClaimSection.args = {
-  ...ClaimDetailsStories.ClaimDetails.args,
-  ...ClaimStatusStories.ClaimStatus.args,
+  detailsContent: ClaimDetailsStories.ClaimDetails.args as ClaimDetailsContent,
+  statusContent: ClaimStatusStories.ClaimStatus.args as ClaimStatusContent,
```
rocketnova, line 18: Note: For stories, I've put test/dummy values into the child component and then referenced those values in parent components. In this case, the default dummy values are in `ClaimDetails.stories.tsx` and they are pulled into `ClaimSection.stories.tsx` here.

stories/ClaimStatus.stories.tsx
```diff
@@ -10,4 +10,7 @@ export default {
 const Template: Story<ClaimStatusProps> = (args) => <ClaimStatusComponent {...args} />
 
 export const ClaimStatus = Template.bind({})
-ClaimStatus.args = {}
+ClaimStatus.args = {
+  statusDescription: 'claim-status:base-pending.description',
```
rocketnova, line 14: Note: I'm wondering if it would be useful to convert this into a radio button or drop down selection in the Storybook. Thoughts?

tests/components/main.test.tsx
```diff
@@ -8,14 +10,27 @@ jest.mock('next/router', () => ({
   useRouter: jest.fn(),
 }))
 
+let scenarioContent: ScenarioContent
+
+beforeAll(() => {
+  const pendingDeterminationScenario = { pendingDetermination: ['temporary text'] }
+  scenarioContent = getScenarioContent(pendingDeterminationScenario)
+})
```
rocketnova, line 18: @lomky @kalvinwang  This chunk of code is reused in both `main.test.tsx` and `index.test.tsx`. Any ideas for where it could go to be DRY-er?

utils/getScenarioContent.tsx
```diff
@@ -0,0 +1,98 @@
...
+  // and hasPendingWeeks is false
+  else if (claimData.hasPendingWeeks === false) {
+    return ScenarioType.BaseNoPending
+  }
+  // This is unexpected
+  // @TODO: Log the scenario and display 500
+  else {
+    // throw new Error('Unexpected scenario')
+    return ScenarioType.BaseNoPending
+  }
```
rocketnova, line 37: Note: This is temporary. In prod, this should result in 500, but I haven't yet resolved how we want to handle this during testing/development.

pages/index.tsx
```diff
@@ -30,9 +32,13 @@ export default function Home({ claimData, loading }: HomeProps): ReactElement {
       </Head>
       <WorkInProgress />
       <Header />
-      <Main loading={loading} />
+      <Main
+        loading={loading}
+        statusContent={scenarioContent.statusContent}
+        detailsContent={scenarioContent.detailsContent}
```
rocketnova, line 38: Note: I broke this out into two arguments, rather than one. I don't feel super strongly about this, so let me know if anyone would prefer that we pass one object instead.

types/common.tsx
```diff
@@ -0,0 +1,40 @@
...
+  claimBalance: string
+  weeklyBenefitAmount: string
+  lastPaymentIssued: string
+  extensionType: string
+}
+
+export interface ScenarioContent {
+  statusContent: ClaimStatusContent
+  detailsContent: ClaimDetailsContent
+}
```
rocketnova, line 40: Note: I'm not attached to these names. Please chime in if you have suggestions/preferences.

tests/utils/getScenarioContent.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const basePending: ScenarioContent = getScenarioContent(basePendingScenario)
+    expect(basePending.statusContent.statusDescription).toBe('claim-status:base-pending.description')
+
+    const baseNoPending: ScenarioContent = getScenarioContent(baseNoPendingScenario)
+    expect(baseNoPending.statusContent.statusDescription).toBe('claim-status:base-no-pending.description')
+  })
+})
+
+// Test getScenario()
+describe('Identifying the scenario', () => {
```
lomky, line 28: Could we add groupings based on the expected results? Clustering like that will help me parse all the tests more cleanly

lomky (commented):
This is looking so good! I'm so excited to have behavior code. Buncha commented embedded!

types/common.tsx
rocketnova: Sounds good. I'll add something and let me know what you think

utils/getScenarioContent.tsx
```diff
@@ -0,0 +1,98 @@
...
+ * Identify the correct scenario to display.
+ *
+ * @param {Qbject} claim
+ * @returns {Object}
+ */
+export function getScenario(claimData: Claim): ScenarioType {
+  // The pending determination scenario: if claimData contains any pendingDetermination
+  // objects
+  // @TODO: refactor with more detailed pending determination scenarios #252
+  if (claimData.pendingDetermination && claimData.pendingDetermination.length > 0) {
```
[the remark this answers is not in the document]
rocketnova: If `pendingDetermination` is `false` and we use `if (claimData.pendingDetermination?.length > 0)`, that should return `false` right?

utils/queryApiGateway.tsx
```diff
@@ -2,11 +2,8 @@ import path from 'path'
 import fs from 'fs'
 import https from 'https'
 import fetch, { Response } from 'node-fetch'
-import type { IncomingMessage } from 'http'
```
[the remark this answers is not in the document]
rocketnova: It wasn't necessary (only [necessary in some circumstances](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#which-should-i-use), so I removed it.

stories/ClaimStatus.stories.tsx
lomky: I think it sounds nice, if its not a huge lift!

utils/getScenarioContent.tsx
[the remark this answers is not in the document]
lomky: it would not - `.?` only handles nullish (`null` or `undefined`) so if we think it could come back as `false` we would want to leave the longer version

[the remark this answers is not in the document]
rocketnova: Ah I see. Then I think we should leave the longer version.

tests/components/main.test.tsx
lomky: If we want to do minimal worrying on refactoring here while we're focusing on business logic, I think my simplest suggestion for drying the tests up would be to just drop the redundant test! Once we have a stable first pass of business logic we can revisit testing the component layers and we'll have more perspective on where calls might go best.

rocketnova: Sounds good. I'll do that for this PR. 

For our future conversation: After talking it through with @kalvinwang, I think I'm leaning towards keeping the `Main` component and moving the business logic into that layer.

lomky (approved):
LGTM! I suggest you updated the PR name / add a high level note in the squash merge so we can ID what point this gets us to easily. :shipit: 🎉 🥇

PR #257, opened 2021-06-24
Add reusable wrapper component around <Trans>.

rocketnova:
## Changes

- Added a new component `<TransLine>`
- Removes router mock code in the header test

## Context

The proposed translation strings contain embedded links. To translate them using [next-i18next](https://github.com/isaachinman/next-i18next), we'll use the [react-i18n `<Trans>`](https://react.i18next.com/latest/trans-component) component. 

This PR introduces a reusable component that will build a `<Trans>` component with arbitrary embedded links. As with the react-i18n `<Trans>` component, some care needs to be taken to make sure that the correct replacement tag (e.g. `<0></0>`) is in the translation string that's passed as the `i18nString` argument.

## Testing Instructions

`yarn test`

Use `yarn storybook` to play with the component. It's listed under Atoms > Trans Line. You can edit `common.json` to add additional strings that have multiple links.

### DIFF
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -1,0 +1,21 @@
+import { Trans } from 'react-i18next'
+import React from 'react'
+
+export interface TransLineProps {
+  i18nString: string // Expects an i18n string
+  links?: string[] | null
+}
+
+export const TransLine: React.FC<TransLineProps> = ({ i18nString, links = null }) => {
+  let linkComponents: JSX.Element[] = []
+  if (links && links.length > 0) {
+    // Disabling some linting rules for this line. The anchor <a> element will
+    // be interporlated by <Trans>.
+    /* eslint-disable jsx-a11y/anchor-has-content */
+    /* eslint-disable react/self-closing-comp */
+    linkComponents = links.map((link, index) => <a href={link} key={index}></a>)
+    /* eslint-enable jsx-a11y/anchor-has-content */
+    /* eslint-enable react/self-closing-comp */
+  }
+  return <Trans i18nKey={i18nString}>{linkComponents}</Trans>
+}
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -13,7 +13,7 @@ export const TransLine: React.FC<TransLineProps> = ({ i18nString, links = null }
     // be interporlated by <Trans>.
     /* eslint-disable jsx-a11y/anchor-has-content */
     /* eslint-disable react/self-closing-comp */
-    linkComponents = links.map((link, index) => <a href={link} key={index}></a>)
+    linkComponents = links.map((link) => <a href={link} key={link}></a>)
     /* eslint-enable jsx-a11y/anchor-has-content */
     /* eslint-enable react/self-closing-comp */
   }
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -2,11 +2,11 @@ import { Trans } from 'react-i18next'
 import React from 'react'
 
 export interface TransLineProps {
-  i18nString: string // Expects an i18n string
+  i18nKey: string // Expects an i18n key
   links?: string[] | null
 }
 
-export const TransLine: React.FC<TransLineProps> = ({ i18nString, links = null }) => {
+export const TransLine: React.FC<TransLineProps> = ({ i18nKey, links = null }) => {
   let linkComponents: JSX.Element[] = []
   if (links && links.length > 0) {
     // Disabling some linting rules for this line. The anchor <a> element will
@@ -17,5 +17,5 @@ export const TransLine: React.FC<TransLineProps> = ({ i18nString, links = null }
     /* eslint-enable jsx-a11y/anchor-has-content */
     /* eslint-enable react/self-closing-comp */
   }
-  return <Trans i18nKey={i18nString}>{linkComponents}</Trans>
+  return <Trans i18nKey={i18nKey}>{linkComponents}</Trans>
 }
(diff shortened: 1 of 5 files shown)


rocketnova:
This was complicated enough that I think some writing some tests for it would be worthwhile. I'd love to pair with someone on that next week if anyone is up for it.

stories/TransLine.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  title: 'Component/Atoms/Trans Line',
+  component: TransLineComponent,
+} as Meta
+
+const Template: Story<TransLineProps> = (args) => <TransLineComponent {...args} />
+
+export const TransLine = Template.bind({})
+TransLine.args = {
+  i18nString: 'work-in-progress.message',
+  links: ['https://navapbc.com', 'https://edd.ca.gov'],
```
lomky, line 14: I'm a little confused by what's passed in vs what I see here!

 - input: i18nstring with one link substition, two links
 - output: text string with one line, the _latter_ one

is that expected behavior to get the second link/last link? why are we passing in two?

lomky: ahah, its taking the link from the index referenced in the key (`1`), so that makes sense there. Is the reason we're passing in two links just to get the positional syntax correct for the i18n string? I think only passing in one item and correcting the index might be cleaner

components/TransLine.tsx
```diff
@@ -0,0 +1,21 @@
+import { Trans } from 'react-i18next'
+import React from 'react'
+
+export interface TransLineProps {
+  i18nString: string // Expects an i18n string
+  links?: string[] | null
```
lomky, line 6: I did a quick test, and I don't think this actually supports null `links` - I get out `undefined` instead of the string for the key I pass in. Any reason not to hard require the links component?

components/TransLine.tsx
```diff
@@ -0,0 +1,21 @@
+import { Trans } from 'react-i18next'
+import React from 'react'
+
+export interface TransLineProps {
+  i18nString: string // Expects an i18n string
```
lomky, line 5: Nitpick: what do you think about calling this `i18nKey` instead? At first blush I don't think I know which side of the object `string` refers to.

stories/TransLine.stories.tsx
rocketnova: > ahah, its taking the link from the index referenced in the key (`1`), so that makes sense there. Is the reason we're passing in two links just to get the positional syntax correct for the i18n string? 

Yes, that's correct. 

> I think only passing in one item and correcting the index might be cleaner

I agree, but I didn't want to introduce a translation string just for the sake of a story, which is why I re-used `work-in-progress.message`. I'm not sure what is the best path forward here. Thoughts?

components/TransLine.tsx
rocketnova: 🤔 Interesting. I'll take another look at this!

rocketnova: Hmm maybe you could show me what you're seeing? I'm not able to reproduce what you described. 

When I do this:

```react
<TransLine i18nKey="header.alt-image-edd" />
```
I see the full translated string like so (not a super helpful screenshot because it's on white background with no context):
Screen Shot 2021-06-28 at 1 25 26 PM

I'd prefer not to require the `links` argument because I want to be able to use the `<TransLine>` component for strings that don't have links, but do have `<strong>` or other more complicated translation elements.

stories/TransLine.stories.tsx
lomky: How about an explainer comment in the code, and maybe name the first URL something obviously placeholder-y, like `example.com/spacer`?

components/TransLine.tsx
lomky: Ah I was testing via adding another story rather than calling the component 
```
 export const NoLinks = Template.bind({})                                                                                
 TransLine.args = {                                                                                                      
   i18nKey: 'title',                                                                                                     
 } 
```   
image

lomky: aaaand I spotted my own bug. I didnt rename the variable I was setting args on - it works now!

tests/components/header.test.tsx
```diff
@@ -1,20 +1,8 @@
 import { render, screen } from '@testing-library/react'
 import { Header } from '../../components/Header'
```
lomky, line 3: ty ty 🙇🏼

setupTests.tsx
```diff
@@ -14,10 +14,24 @@ import enClaimStatus from './public/locales/en/claim-status.json'
...
+      test: {
+        transLine: {
+          plainString: 'just text',
+          plainStringOneLink: 'first <0>second</0> third',
+          plainStringLinks: 'first <0>second</0> <1>third</1>',
+          plainStringLinksComplicated: '<1>first</1> <0>second</0> third <0>fourth</0> <1>fifth</1>',
+          styledString: 'first <strong>second</strong> third',
+          styledStringOneLink: 'first <strong>second</strong> <0>third</0>',
+          styledLink: 'first <strong><0>second</0></strong>',
+        },
```
lomky, line 32: I really find myself wanting these strings defined closer to the tests. Let me see if I can find anything in a bit of searching

lomky: it looks like this comes down to mocking, which I don't have a great grasp of, so I'll but this aside for now

lomky (approved):
🎏 :shipit:

setupTests.tsx
rocketnova: Ok sounds good. Happy to revisit anytime.

PR #262, opened 2021-06-29
Request converts header keys to lowercase

rocketnova:
## Changes

- Breaks out retrieving the unique number into a separate helper function for easier testing
- Converts the `ID_HEADER_NAME` env var to lowercase

## Context

I discovered that the `req` argument in `GetServerSideProps()` passes the header keys as lowercase (e.g. `user-agent` instead of `User-Agent`). To be able to retrieve the unique number correctly, we need to convert the value of `ID_HEADER_NAME` env var to lowercase as well.

## Testing Instructions

`yarn test`


lomky (approved):
:shipit:

Issue #123, opened 2021-04-23
Update default Next.js error pages


kalvinwang:
### Description

- If there's an issue with the Claim Tracker web application itself, we should show the appropriate error page (404 page not found or 5xx server-side error)
- Next.js has default 404 and 500 error pages (see below); we can [customize](https://nextjs.org/docs/advanced-features/custom-error-page) them.
- The Claim Tracker 404/500 error pages should look obviously different from the UIO 404/500 error pages, s that when a user who takes a screenshot of the page, it's clear whether the error was in UIO vs Claim Tracker. The default Next.js error pages are sufficiently different already. However, we may want to customize the Next.js error pages to look more like part of UIO than they do by default.

Context:
-  If there's an issue with the IDM integration, the Claim Tracker page won't be loaded, and an IDM error page will be shown, hopefully
- If there's an issue reaching the API gateway, we'll return some kind of inline error, TBD in a separate ticket.

Default 404 page:
![image](https://user-images.githubusercontent.com/82277/116940035-4d02a900-ac3b-11eb-863f-f36fbf4f9747.png)

### Acceptance Criteria

  - [ ] On staging, should show the appropriate error page (404 or 5xx) when triggered

nguyen-tina:
Mocked up new 404 design and posted to channel for review

nguyen-tina:
Updated [proposal](https://docs.google.com/document/d/1SX5-cKj10MdjMqdacHAsKZyasd2vaERWSRTCfPL3-J4/edit#heading=h.qri95rtk659n) for 404 and 500 content. Handing off to Engineering.

PR #263, opened 2021-06-29
Use default nextjs error page

rocketnova:
## Changes

- Display default next.js error page if an error is encountered when querying the API gateway

## Context

For #242, I want to be able to throw errors and have them caught & display a 500 error (we may want to display 404 in some cases, but I'm not sure which ones yet).

I attempted to move the [`getServerSideProps()`](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering) logic out of index.tsx and into main.tsx. However, I discovered that you can only use it in [Page level](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering) components. In light of this, I have [changed my mind](https://github.com/cagov/ui-claim-tracker/pull/256#discussion_r658328707) and think that we should collapse main.tsx directly into index.tsx (not included in this PR). Let me know if this sounds good and I can make a ticket for it/feel free to make a ticket for it.

Note that we have a separate ticket (#123) for customizing the default next.js error pages.

## Test

`yarn storybook`

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -79,6 +79,8 @@ export const getServerSideProps: GetServerSideProps = async ({ req, locale }) =>
     // Run business logic to get content for the current scenario.
     scenarioContent = getScenarioContent(claimData)
   } catch (error) {
+    // If an error occurs, log it and show 500.
+    logger.error(error)
     errorCode = 500
   }
 
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -5,6 +5,7 @@ import { ReactElement } from 'react'
 import { useTranslation } from 'next-i18next'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import { GetServerSideProps } from 'next'
+import Error from 'next/error'
 
 import { Header } from '../components/Header'
 import { Main } from '../components/Main'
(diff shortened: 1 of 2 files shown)


lomky:
> I attempted to move the getServerSideProps() logic out of index.tsx and into main.tsx. However, I discovered that you can only use it in Page level components. In light of this, I have changed my mind and think that we should collapse main.tsx directly into index.tsx (not included in this PR). Let me know if this sounds good and I can make a ticket for it/feel free to make a ticket for it.

Sounds right to me. File the issue and let me know if you'd like to to pick it up so you can stay focused!

pages/index.tsx
```diff
@@ -55,17 +68,28 @@ export const getServerSideProps: GetServerSideProps = async ({ req, locale }) =>
...
-  const claimData = await queryApiGateway(req)
+  let errorCode: number | null = null
+  let claimData: Claim | null = null
+  let scenarioContent: ScenarioContent | null = null
 
-  // Run business logic to get content for the current scenario.
-  const scenarioContent = getScenarioContent(claimData)
+  try {
+    // Make the API request and return the data.
+    claimData = await queryApiGateway(req)
```
lomky, line 77: Nit: Any reason not to declare `claimData` here instead of the outer scope?

rocketnova: Good catch! Pushed a commit to correct this.

rocketnova:
> > I attempted to move the getServerSideProps() logic out of index.tsx and into main.tsx. However, I discovered that you can only use it in Page level components. In light of this, I have changed my mind and think that we should collapse main.tsx directly into index.tsx (not included in this PR). Let me know if this sounds good and I can make a ticket for it/feel free to make a ticket for it.
> 
> Sounds right to me. File the issue and let me know if you'd like to to pick it up so you can stay focused!

Opened #265. Would you mind picking it up?

lomky (approved):

Issue #150, opened 2021-05-06
Validate API Gateway response


kalvinwang:
### Description
We should validate each field in the API Gateway (once we find out what those fields are). Consider using Joi, or creating a separate ticket for that if it seems unnecessary now but helpful later.

### Acceptance Criteria

  - [ ] Get possible values for each field
  - [ ] Write validation code

PR #266, opened 2021-06-29
Split scenario "base state (no pending weeks)" into two sub scenarios

rocketnova:
## Ticket

Related to #242 

## Changes

- Splits the Base State scenarios into the current 4 known scenarios (more refinement TBD)
- Updates scenarios to be numbered
- Converts scenario type enum from an enum with string members to an enum with number members
- Adds a type alias for i18n translation strings so that we can use the type `I18nString` instead of just `string` for extra clarity

## Context

While going through the content matrix, I discovered that I didn't break out the scenarios at the correct level of granularity in order to display the content we want to show. After getting confirmation about this from @nicoleslaw and Diana, this PR corrects that.

## Testing 

`yarn test`


public/locales/en/claim-status.json
```diff
@@ -3,9 +3,18 @@
...
+  "base-no-pending-active": {
+    "description": "Your claim is active and there are no pending issues at this time."
+  },
+  "base-no-pending-inactive": {
     "description": "There are no pending issues at this time."
+  },
+  "conditional-next-steps": {
+    "certify-no-pending": "<0>Keep certifying</0> as long as you need benefits.",
+    "certify-pending": "<0>Keep certifying</0> as long as you need benefits, so your payments are not delayed further.",
+    "contact-info": "Make sure we have your current phone number and email address."
```
rocketnova, line 18: Note: This PR pulls in some content that I'm not yet using in this PR.

utils/getScenarioContent.tsx
```diff
@@ -3,14 +3,12 @@ import { Claim, ClaimDetailsContent, ClaimStatusContent, ScenarioContent } from
 export enum ScenarioType {
   PendingDetermination = 'Pending determination scenario',
   BasePending = 'Base state with pending weeks',
-  BaseNoPending = 'Base state with no pending weeks',
+  BaseNoPendingActive = 'Base state with no pending weeks; Active claim',
+  BaseNoPendingInactive = 'Base state with no pending weeks; Inactive claim',
```
lomky, line 7: I find myself really torn on staying brief and wanting more clarity. 

Pending is overloaded and applies to both Determinations and Weeks, so should we always include the associated Noun?

Active/Inactive apply the a whole other object (the claim), should we include that?

But then we have names like:
```
  BaseNoPendingWeeksActiveClaim 
  BaseNoPendingWeeksInactiveClaim 
```
is that just getting absurd and hard to parse?

rocketnova: Totally hear you. I'll take another pass at this.

stories/Page.stories.tsx
```diff
@@ -12,24 +13,23 @@ export default {
   decorators: [withNextRouter],
   argTypes: {
     scenario: {
-      options: Object.keys(ScenarioType),
-      mapping: Object.keys(ScenarioType), // return the key instead of the value
```
rocketnova, line 16: Note: Turns out `mapping` was not needed, so I removed it.

tests/utils/apiGatewayStub.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+    expect(response.hasCertificationWeeksAvailable).toBe(false)
+  })
+
+  it('is correct for Scenario 10', () => {
+    const response = apiGatewayStub(ScenarioType.Scenario10)
+    expect([null, [], false, undefined]).toContain(response.pendingDetermination)
+    expect(response.hasPendingWeeks).toBe(true)
+    expect(response.hasCertificationWeeksAvailable).toBe(true)
+  })
+})
```
rocketnova, line 39: Note: The tests for scenarios 7-10 are pretty repetitive, but I thought it would help readability to have them broken out separately. Also I think these might change soon. Let me know if it would be better to make this one test that verifies against a lookup table instead.

tests/utils/getScenarioContent.test.tsx
```diff
@@ -1,98 +1,76 @@
...
+  it('are returned as expected', () => {
+    const baseScenarios = [
+      ScenarioType.Scenario7,
+      ScenarioType.Scenario8,
+      ScenarioType.Scenario9,
+      ScenarioType.Scenario10,
+    ]
+    for (const scenarioType of baseScenarios) {
+      expect(getScenario(apiGatewayStub(scenarioType))).toBe(scenarioType)
+    }
```
rocketnova, line 74: Note: This is a basically just testing that `getScenario()` can process the output for our stub, so I added tests for our stub.

utils/getScenarioContent.tsx
```diff
@@ -1,75 +1,86 @@
...
+  Scenario9,
+  Scenario10,
+}
+
+export const ScenarioTypeNames = {
+  [ScenarioType.Scenario1]: 'Pending determination scenario',
+  [ScenarioType.Scenario7]: 'Base state; No pending weeks; No weeks to certify',
+  [ScenarioType.Scenario8]: 'Base state; No pending weeks; Has weeks to certify',
+  [ScenarioType.Scenario9]: 'Base state; Has pending weeks; No weeks to certify',
+  [ScenarioType.Scenario10]: 'Base state; Has pending weeks; Has weeks to certify',
```
rocketnova, line 16: Note: This is the major change. After speaking with Nicole and Diana today, we agreed to call scenarios by number instead of a nick name. As part of that, I converted the enum to a numeric enum, which is able to do automatic [reverse mappings](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings). Enums are a little finicky with Typescript, so I added the `numericEnum.tsx` utility to help.

utils/getScenarioContent.tsx
```diff
@@ -1,75 +1,86 @@
...
-  // and hasPendingWeeks is false
-  else if (claimData.hasPendingWeeks === false) {
-    return ScenarioType.BaseNoPending
-  }
-  // This is unexpected
-  // @TODO: Log the scenario and display 500
+
+  // Otherwise display a Base State scenario.
   else {
-    // throw new Error('Unexpected scenario')
```
rocketnova, line 37: Note: The new logic I went over with Nicole and Diana shouldn't leave an opportunity for an unknown scenario. We will also have an additional layer of API gateway response validation (#150) that will make unexpected response values not possible.

utils/getScenarioContent.tsx
lomky: I like it. Can we add a comment to the code that points to the fully-written-out Scenario descriptions?

rocketnova: @lomky Yes for sure. But I'm not sure exactly where you are thinking we should add the comment? In `getScenarioContent.tsx`?

lomky: I was thinking right above this const, or if we do topline comments on the files, there would also be good!

lomky (approved):
🎉

PR #269, opened 2021-06-29
Update README.md

kalvinwang:
We've already been running Node 14 LTS in Azure on all environments, and our code uses some JS syntax that landed in Node 14


rocketnova (approved):
🚢

PR #274, opened 2021-07-01
Add complex value to the Page story

rocketnova:
## Changes

- Adds the ability to select a scenario in the `Page` story to display all content that will be displayed for that scenario (currently only the Claim Status description)
- Removes the JSON blob for `scenarioContent` on the `Page` story (causes conflicts)
- Adds`apiGatewayStub` function to stub responses from the API gateway (now needed for both tests and stories)
- Builds out the `Claim` type a little bit more

## Context

I'm getting close to the point where I'm going to stop displaying placeholder content when the request doesn't contain the unique number. So I need a way to view the content for each scenario. 

This PR sets the foundation for more complex story controls. In a near-future PR, I'll likely extend this foundation so that we can also specify whether there are certification weeks available (which impacts what Next Steps are shown).

## Testing

`yarn storybook`

### DIFF
--- a/stories/Page.stories.tsx
+++ b/stories/Page.stories.tsx
@@ -15,6 +15,7 @@ export default {
       type: { name: ScenarioType, required: false },
       options: Object.keys(ScenarioType),
       mapping: Object.keys(ScenarioType), // return the key instead of the value
+      defaultValue: 'BaseNoPending',
       control: {
         type: 'select',
         labels: ScenarioType,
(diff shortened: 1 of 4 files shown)


stories/Page.stories.tsx
```diff
@@ -2,19 +2,38 @@ import { Story, Meta } from '@storybook/react'
...
+    scenario: {
+      options: Object.keys(ScenarioType),
+      mapping: Object.keys(ScenarioType), // return the key instead of the value
+      defaultValue: 'BaseNoPending',
+      control: {
+        type: 'select',
+        labels: ScenarioType,
+      },
+    },
+  },
```
lomky, line 23: This is so cool!

lomky (approved):
🎉 🎉 🎉

Issue #218, opened 2021-06-07
Document how we want to trigger alerts


kalvinwang:
### Description
How will alerts be triggered?

### Acceptance Criteria

  - [ ] Document how we'll trigger alerts

rocketnova:
I've updated the Foundations document with text that is ready for review.

rocketnova:
I've returned to this ticket and re-updated the Foundations doc with the latest changes.

rocketnova:
I'm closing this ticket. We discussed the changes to the Foundations doc and agreed on them in Eng sync yesterday.

PR #288, opened 2021-07-07
Configure streaming to Application Insights 

rocketnova:
## Ticket

Related to #218 

## Changes

- Adds [`pino-applicationinsights`](https://github.com/ovhemert/pino-applicationinsights) as a dependency to log directly to Azure Application Insights in Production
- Adds a type declaration file for `pino-applicationinsights`

## Context

To facilitate setting up and tuning alerts, I'm using `pino-applicationinsights`'s [`createWriteStream()`](https://github.com/ovhemert/pino-applicationinsights/blob/master/docs/API.md) for Production environments (i.e. `process.env.NODE_ENV === 'production'`). In non-prod environments, `pino` will work as before (outputting to STDOUT and using pretty print).

Because `pino` is able to take a [WritableStream as a `destination`](https://github.com/pinojs/pino/blob/master/docs/api.md#destination), I didn't need to use [`pino-multi-stream`](https://github.com/pinojs/pino-multi-stream).

Note: This change means that in Production environments, we will no longer see the output of any `logger.whatever()` calls in the Azure App Services App Stream or Log Analytics. Instead, the output will go directly to the associated Azure Application Insights Logs (in the `traces` table).

Because `pino-applicationinsights` is basically a very nice wrapper around [`applicationinsights`](https://github.com/microsoft/ApplicationInsights-node.js#readme), it needs an Application Insights instrumentation key in order to send logs to the write App Insights. By default, `pino-applicationinsights` knows to look for `process.env.APPINSIGHTS_INSTRUMENTATIONKEY`.

### Type Declaration

Because `pino-applicationinsights` doesn't have a type declaration and none exists for this package in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/), I created a rough one. I've opened #289 to polish it up and submit a PR to DefinitelyTyped. 

## Testing

1. Run a pipeline for this branch
2. Release the pipeline on a non-prod environment 
3. Navigate to the claim tracker URL in the non-prod environment
4. Trigger an error (will throw 500s right now because of unrelated certificate issue)
5. Navigate to Application Insights (not App Services!) in matching non-prod environment
6. View Logs and query for `traces` (may take a couple minutes for the logs to populate)
7. You should see a SEV3 with the message "Application error" for the timestamp that you loaded the page:
Screen Shot 2021-07-07 at 4 40 13 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "next": "10.0.8",
     "node-fetch": "^2.6.1",
     "pino": "^6.11.3",
+    "pino-applicationinsights": "^2.1.0",
     "react": "17.0.1",
     "react-bootstrap": "^1.5.2",
     "react-dom": "17.0.1"
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@testing-library/react": "^11.2.5",
     "@types/node": "^14.14.33",
     "@types/pem": "^1.9.5",
+    "@types/pumpify": "^1.4.1",
     "@types/react": "^17.0.3",
     "@types/react-test-renderer": "^17.0.1",
     "@typescript-eslint/eslint-plugin": "^4.19.0",
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -86,13 +86,13 @@ export default function Home({
 }
 
 export const getServerSideProps: GetServerSideProps = async ({ req, locale }) => {
-  const isProd = process.env.NODE_ENV === 'production'
+  const isAzureEnv = process.env.NODE_ENV === 'production'
 
-  // Pino: use pretty print and log to STDOUT if not in production mode.
+  // Pino: use pretty print and log to STDOUT for local environments.
   let logger = pino({ prettyPrint: true })
 
   // Pino: otherwise, log to Azure Application Insights.
-  if (isProd) {
+  if (isAzureEnv) {
     const appInsightsStream = await appInsights.createWriteStream()
     logger = pino(appInsightsStream)
   }
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,6 +1,8 @@
 import Head from 'next/head'
 import Container from 'react-bootstrap/Container'
 import pino from 'pino'
+import appInsights from 'pino-applicationinsights'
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -54,6 +55,7 @@
     "@testing-library/react": "^11.2.5",
     "@types/node": "^14.14.33",
     "@types/pem": "^1.9.5",
+    "@types/pumpify": "^1.4.1",
```
rocketnova, line 58: Note: These types are needed for the type definition for `pino-applicationinsights`

pages/index.tsx
```diff
@@ -100,7 +110,7 @@ export const getServerSideProps: GetServerSideProps = async ({ req, locale }) =>
     scenarioContent = getScenarioContent(claimData)
   } catch (error) {
     // If an error occurs, log it and show 500.
-    logger.error(error)
+    logger.error(error, 'Application error')
```
rocketnova, line 113: Note: I added a message string to make it easier to locate these as new messages in Application Insights. We can change this at any time.

tsconfig.json
```diff
@@ -18,12 +14,6 @@
...
   },
-  "include": [
-    "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx"
-  ],
-  "exclude": [
-    "node_modules"
-  ]
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "types"],
```
rocketnova, line 17: Note: Added `types` so that Typescript would find the new type definitions. All other changes in this file are auto-formatting.

lomky (approved):
🎉

pages/index.tsx
```diff
@@ -86,7 +87,16 @@ export default function Home({
 
 export const getServerSideProps: GetServerSideProps = async ({ req, locale }) => {
   const isProd = process.env.NODE_ENV === 'production'
```
kalvinwang, line 89: what do you think about replacing `isProd` with something that isn't ambiguous? since test, staging, etc Azure envs are all "isProd". something like `isAzureEnv` or `isNotLocalEnv`

kalvinwang (approved):
lgtm! one small suggestion

PR #293, opened 2021-07-08
Remove dotenv package

### LINKED ISSUE #292: Use built-in Next.js environment variable support
### Description

Next.js has built-in support for [environment variables](https://nextjs.org/docs/basic-features/environment-variables). Let's  use that and reduce the number of dependencies.

### Acceptance Criteria

- [ ] Remove [dotenv](https://www.npmjs.com/package/dotenv) dependency

rocketnova:
## Ticket

Closes #292 

## Changes

- Removes [dotenv](https://github.com/motdotla/dotenv) as a dependency 

## Context

Next.js has built-in support for [environment variables](https://nextjs.org/docs/basic-features/environment-variables), so we don't need to use a separate package for that.

⚠️ This requires a change for local development: **Rename .env to .env.local** ⚠️ 

Technically, Next.js will still load `.env`, but since `.env.local` is safe for secrets, we should use that.

## Testing

1. Check out this branch
2. Rename your `.env` file to `.env.local` 
3. `yarn dev`
3. Confirm your env vars are still loading (no error: `Error: EISDIR: illegal operation on a directory, read`)


lomky (approved):
Nice!

kalvinwang (approved):
lgtm, also update the .env file in 1Password to .env.local (and it says "is loaded using dotenv npm package" right now)

PR #307, opened 2021-07-12
Implement logic for Claim Details

### LINKED ISSUE #290: Implement logic for claim details
### Description

Splitting this ticket out from #242. We need some logic to display claim details, including the Program Type and Extension Type names.

### Acceptance Criteria

- [ ] Claim Details are displayed based on information from the API gateway query

rocketnova:
## Ticket

Closes #290 

## Changes

- Adds a `getClaimDetails.tsx` utility to handle Claim Details
- Adds `claim-details.json` translation files
- Adds a `urls.json` file for urls displayed in the content
- Hides extension type field label if there is no extension type
- Installs [date-fns](https://date-fns.org/) for date formatting
- Adds claim details to `apiGatewayStub`

## Context

This PR adds functionality to display Claim Details content based on the response from the API gateway. It includes logic to map the program type string returned from the API gateway to user-facing translation strings (including splitting into the Program Type and Extension Type fields).

This PR wires up the new logic to the `<ClaimDetails>` component and does currency and date formatting based on the sample data we have received from the API gateway.

## Testing

`yarn test`


components/ClaimDetails.tsx
```diff
@@ -15,7 +15,17 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
...
 }) => {
-  const { t } = useTranslation('common')
+  const { t } = useTranslation(['common', 'claim-status'])
+
+  let extension: JSX.Element | null = null
+  if (extensionType) {
+    extension = (
+      <div className="col-6">
+        <div />
+        <InfoField loading={loading} label={t('claim-details.extension-type')} text={t(extensionType)} />
```
lomky, line 25: Wait how do these work?! You imported `claim-status` (presumably a bug), but the translations are still working!

rocketnova: Whoa! That was a bug. I meant to import `claim-details` but maybe it doesn't need to explicitly import here because it's imported in `index.tsx`...?

lomky: ooo maybe! we do need to get access to `t` here, but maybe there's some trick

utils/getClaimDetails.tsx
```diff
@@ -0,0 +1,129 @@
...
+  // Get programType and extensionType.
+  const pair: programExtensionPairType = getProgramExtensionPair(rawDetails.programType)
+  const benefitYear = buildBenefitYear(rawDetails.benefitYearStartDate, rawDetails.benefitYearEndDate)
+
+  return {
+    programType: pair.programType,
+    benefitYear: benefitYear,
+    claimBalance: formatCurrency(rawDetails.claimBalance),
+    weeklyBenefitAmount: formatCurrency(rawDetails.weeklyBenefitAmount),
+    lastPaymentIssued: formatDate(rawDetails.lastPaymentIssued),
```
lomky, line 126: Should we also be pulling out the `lastPaymentAmount`?

utils/getClaimDetails.tsx
```diff
@@ -0,0 +1,129 @@
...
+  throw new Error('Unknown Program Type')
+}
+
+/**
+ * Format dates.
+ */
+export function formatDate(dateString: string): string {
+  const parsedDate = parse(dateString, "yyyy-MM-dd'T'HH:mm:ss", new Date())
+  return format(parsedDate, 'M/d/yyyy')
+}
```
lomky, line 97: Since we only use this here, its chill, but if we ever need a date in another component scope I think this is a good candidate for its own util!

lomky: (ditto the money one!)

tests/utils/getClaimDetails.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  it('results in a date range string', () => {
+    const range = buildBenefitYear('2013-09-27T12:34:33', '2021-10-28T05:33:34')
+    expect(range).toBe('9/27/2013 - 10/28/2021')
+  })
+})
+
+// Test formatCurrency()
+describe('Formatting dollar amounts', () => {
+  it('displays the correct string for amounts with cents', () => {
+    const withCents = '532.82'
```
lomky, line 48: Should these be numbers, like the function is expecting, or should the function be taking strings and handling the conversion?

lomky: I think you can drop all but one test - we only care about checking 'are we getting the output we expect', we don't have to re-test the underlying function!

tests/utils/getClaimDetails.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    // Mock ClaimDetailsResults
+    const rawDetails = {
+      programType: 'UI',
+      benefitYearStartDate: '2021-05-21T00:00:00',
+      benefitYearEndDate: '2022-05-20T00:00:00',
+      claimBalance: 100,
+      weeklyBenefitAmount: 25,
+      lastPaymentIssued: '2021-03-12T00:00:00',
+      lastPaymentAmount: 10,
+      monetaryStatus: 'active',
```
lomky, line 77: I think I flagged this already, but these two inputs just quietly disappear!

components/ClaimDetails.tsx
```diff
@@ -34,10 +44,7 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
 
             <InfoField loading={loading} label={t('claim-details.last-payment-issued')} text={lastPaymentIssued} />
```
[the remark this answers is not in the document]
rocketnova: We might want to add `lastPaymentAmount` but I don't believe we'll add `monetaryStatus` to this section. I think `monetaryStatus` is only used for determining business logic and not displayed directly.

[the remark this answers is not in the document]
rocketnova: I've asked Tina et al about `lastPaymentAmount` in Slack. Let's leave it out of this PR for now and open a new ticket if need to display it.

utils/getScenarioContent.tsx
```diff
@@ -93,25 +94,10 @@ export default function getScenarioContent(claimData: Claim): ScenarioContent {
...
-    detailsContent.benefitYear = `${claimData.claimDetails.benefitYearStartDate} - ${claimData.claimDetails.benefitYearEndDate}`
-    detailsContent.claimBalance = claimData.claimDetails.claimBalance
-    detailsContent.weeklyBenefitAmount = claimData.claimDetails.weeklyBenefitAmount
-    detailsContent.lastPaymentIssued = claimData.claimDetails.lastPaymentIssued
-    // @TODO
-    // detailsContent.extensionType = ''
+  if (!claimData.claimDetails) {
+    throw new Error('Missing claim details')
   }
+  const detailsContent: ClaimDetailsContent = getClaimDetails(claimData.claimDetails)
```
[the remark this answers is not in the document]
rocketnova: Yeah very fair. I think I've found the distinction to be moot, but I'm happy to adopt more rigor around it. Do you have any suggestions?

utils/getClaimDetails.tsx
rocketnova: See https://github.com/cagov/ui-claim-tracker/pull/307#discussion_r669174902

utils/getClaimDetails.tsx
rocketnova: Yup agreed. I thought of this as the first step before refactoring into another util.

tests/utils/getClaimDetails.test.tsx
rocketnova: See https://github.com/cagov/ui-claim-tracker/pull/307#discussion_r669174902

tests/utils/getClaimDetails.test.tsx
rocketnova: > Should these be numbers, like the function is expecting, or should the function be taking strings and handling the conversion?

Yes! Good catch. I'll fix this.

> I think you can drop all but one test - we only care about checking 'are we getting the output we expect', we don't have to re-test the underlying function!

Fair! I'm always antsy about making sure to test that I've configured the thing to use the underlying function correctly, but I can pare this one down.

components/ClaimDetails.tsx
rocketnova: From https://github.com/isaachinman/next-i18next#4-declaring-namespace-dependencies:

> Note: useTranslation provides namespaces to the component that you use it in. However, serverSideTranslations provides the total available namespaces to the entire React tree and belongs on the page level. Both are required.

In my testing it seems like I can do `const { t } = useTranslation()` in `ClaimDetails.tsx` and the translations actually work fine, but because the documentation says it's required to pass in the namespace to `useTranslation()`, it seems safer to just do that. I'll fix this bug.

utils/getScenarioContent.tsx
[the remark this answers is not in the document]
lomky: I slightly prefer build, as get suggests you might be calling someone. but I'm not super bothered either way

lomky (approved):
I think its still worth doing the get/build names reconciled, but you can totally do that here or as a follow on later, whichever works!

utils/getScenarioContent.tsx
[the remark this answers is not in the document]
rocketnova: 👍 Sounds good. I don't have a preference, so I'll update these to `buildTheThing` in a future PR.

components/ClaimDetails.tsx
rocketnova: I take it back. I'm not sure what was going on with my testing yesterday, but we do need `const { t } = useTranslation(['common', 'claim-details'])`in `ClaimDetails.tsx` for the translation strings to work.

lomky: I've noticed some issues around hot reloading and translation files - might be that!

Issue #299, opened 2021-07-09
Installing pino-applicationinsights has broken storybook builds


rocketnova:
### Description

Sadly, #288 broke storybook builds. See https://github.com/cagov/ui-claim-tracker/runs/3023514358?check_suite_focus=true. `yarn storybook-build` is also broken locally for `main`.

If this takes more than a day or two, consider whether it does make sense to use pino-applicationinsights

### Acceptance Criteria

- [ ] Storybook builds work again

lomky:
Poked at this for a minute - the sb build fails whether isAzureEnv is true or false, and succeeds if you comment out the await createWriteStream, so its something in the actual build step there

PR #314, opened 2021-07-14
Update base state and pending state scenarios

### LINKED ISSUE #291: Update business logic for Base State scenarios
### Description

Splitting this ticket out from #242. We will need to update the business logic for the Base State scenarios once we have more finalized content. 

### Acceptance Criteria

- [ ] Base States are displaying final content

rocketnova:
## Ticket

Resolves #291 

## Changes

- Re-numbers the scenarios and updates claim-status.json based on latest content
- Splits claim status description into two fields: `heading` and `summary` according to content docs
- Uses the `<TransLine>` component because the claim status summary field can contain links
- Introduces looking up urls in `urls.json` by key

## Context

Based on the latest content, this PR replaces Scenarios 7-10 with Scenarios 4-6. The new content introduces a different claim status heading for each scenario, as well as a claim status "summary". The claim status summary text can contain links, so we need to start using `<TransLine>`.

In an effort to make it easier to update content, I'm using the following concepts:
- All urls are stored as key value pairs in `urls.json`
- The English translation file (e.g. `public/locales/en/claim-status.json`) includes an array of keys for `urls.json`

This allows us to:
1. Keep the content together in one document (the translation file)
2. Declare urls once, rather than copy/pasting them multiple times and introducing types

## Testing

`yarn test`

and 

While #299 is not resolved, comment out these lines, then run `yarn storybook`, and check that each scenario displays the heading and summary from the latest content (including correct links): https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98

### DIFF
--- a/components/ClaimStatus.tsx
+++ b/components/ClaimStatus.tsx
@@ -2,13 +2,14 @@ import { useTranslation } from 'next-i18next'
 
 import { NextSteps } from './NextSteps'
 import { TextLine } from './TextLine'
+import { TransLine } from './TransLine'
 import { ClaimStatusContent } from '../types/common'
 
 export interface ClaimStatusProps extends ClaimStatusContent {
   loading: boolean
 }
 
-export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, heading, nextSteps }) => {
+export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, heading, summary, nextSteps }) => {
   const { t } = useTranslation(['common', 'claim-status'])
 
   return (
@@ -16,6 +17,9 @@ export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, headi
       <h2>{t('claim-status.title')}</h2>
       <div className="pending-status">
         <TextLine loading={loading} text={t(heading)} />
+      </div>
+      <div className="summary">
+        <TransLine i18nKey={summary.i18nKey} links={summary.links} />
       </div>
       <div className="status-box">
         <div className="topbar">
--- a/components/ClaimStatus.tsx
+++ b/components/ClaimStatus.tsx
@@ -8,14 +8,14 @@ export interface ClaimStatusProps extends ClaimStatusContent {
   loading: boolean
 }
 
-export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, statusDescription, nextSteps }) => {
+export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, heading, nextSteps }) => {
   const { t } = useTranslation(['common', 'claim-status'])
 
   return (
     <div className="claim-status">
       <h2>{t('claim-status.title')}</h2>
       <div className="pending-status">
-        <TextLine loading={loading} text={t(statusDescription)} />
+        <TextLine loading={loading} text={t(heading)} />
       </div>
       <div className="status-box">
         <div className="topbar">
--- a/tests/utils/apiGatewayStub.test.tsx
+++ b/tests/utils/apiGatewayStub.test.tsx
@@ -8,32 +8,27 @@ describe('The API gateway stub response for the Pending Determination scenario',
   })
 })
 
+describe('The API gateway stub response for the Generic Pending state', () => {
+  it('is correct for Scenario 4', () => {
+    const response = apiGatewayStub(ScenarioType.Scenario4)
+    expect([null, [], false, undefined]).toContain(response.pendingDetermination)
+    expect(response.hasPendingWeeks).toBe(true)
+    expect([true, false]).toContain(response.hasCertificationWeeksAvailable)
+  })
+})
+
 describe('The API gateway stub response for the Base States', () => {
-  it('is correct for Scenario 7', () => {
-    const response = apiGatewayStub(ScenarioType.Scenario7)
+  it('is correct for Scenario 5', () => {
+    const response = apiGatewayStub(ScenarioType.Scenario5)
     expect([null, [], false, undefined]).toContain(response.pendingDetermination)
     expect(response.hasPendingWeeks).toBe(false)
     expect(response.hasCertificationWeeksAvailable).toBe(false)
   })
 
-  it('is correct for Scenario 8', () => {
-    const response = apiGatewayStub(ScenarioType.Scenario8)
+  it('is correct for Scenario 6', () => {
+    const response = apiGatewayStub(ScenarioType.Scenario6)
     expect([null, [], false, undefined]).toContain(response.pendingDetermination)
     expect(response.hasPendingWeeks).toBe(false)
-    expect(response.hasCertificationWeeksAvailable).toBe(true)
-  })
-
-  it('is correct for Scenario 9', () => {
-    const response = apiGatewayStub(ScenarioType.Scenario9)
-    expect([null, [], false, undefined]).toContain(response.pendingDetermination)
-    expect(response.hasPendingWeeks).toBe(true)
-    expect(response.hasCertificationWeeksAvailable).toBe(false)
-  })
-
-  it('is correct for Scenario 10', () => {
-    const response = apiGatewayStub(ScenarioType.Scenario10)
-    expect([null, [], false, undefined]).toContain(response.pendingDetermination)
-    expect(response.hasPendingWeeks).toBe(true)
     expect(response.hasCertificationWeeksAvailable).toBe(true)
   })
 })
--- a/tests/utils/getClaimStatus.test.tsx
+++ b/tests/utils/getClaimStatus.test.tsx
@@ -13,22 +13,21 @@ describe('The Claim Status heading', () => {
 })
 
 // Test getClaimStatusSummary()
+// Note: These tests test against real content
 describe('The Claim Status summary', () => {
-  it('is correct for each scenario', () => {
-    for (const key of getNumericEnumKeys(ScenarioType)) {
-      const summary = getClaimStatusSummary(key)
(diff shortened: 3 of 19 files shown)


components/ClaimStatus.tsx
```diff
@@ -2,20 +2,24 @@ import { useTranslation } from 'next-i18next'
...
 
   return (
     <div className="claim-status">
       <h2>{t('claim-status.title')}</h2>
       <div className="pending-status">
-        <TextLine loading={loading} text={t(statusDescription)} />
+        <TextLine loading={loading} text={t(heading)} />
+      </div>
+      <div className="summary">
+        <TransLine i18nKey={summary.i18nKey} links={summary.links} />
```
rocketnova, line 22: @lomky I just tossed the heading and summary in here like this for now and assumed that you'd rework how they get rendered.

utils/getClaimStatus.tsx
```diff
@@ -2,22 +2,65 @@
...
+ * Get Claim Status heading.
+ */
+export function getClaimStatusHeading(scenarioType: ScenarioType): I18nString {
+  return getTranslationPrefix(scenarioType) + '.heading'
+}
+
+/**
+ * Get Claim Status summary.
+ *
+ * Note: This function has a lot of verbosity due to Typescript + eslint.
```
rocketnova, line 27: Note: There's a lot going on in here to make the Typescript work properly.

utils/getScenarioContent.tsx
```diff
@@ -13,18 +13,16 @@ import getClaimStatus from './getClaimStatus'
...
 
 export const ScenarioTypeNames = {
   [ScenarioType.Scenario1]: 'Pending determination scenario',
-  [ScenarioType.Scenario7]: 'Base state; No pending weeks; No weeks to certify',
-  [ScenarioType.Scenario8]: 'Base state; No pending weeks; Has weeks to certify',
-  [ScenarioType.Scenario9]: 'Base state; Has pending weeks; No weeks to certify',
-  [ScenarioType.Scenario10]: 'Base state; Has pending weeks; Has weeks to certify',
+  [ScenarioType.Scenario4]: 'Generic pending state: pending weeks',
+  [ScenarioType.Scenario5]: 'Base state: no pending weeks, no weeks to certify',
+  [ScenarioType.Scenario6]: 'Base state: no pending weeks, weeks to certify',
```
rocketnova, line 25: Note: New scenario numbers!

tests/utils/getClaimStatus.test.tsx
```diff
@@ -1,14 +1,32 @@
...
-      expect(getClaimStatusDescription(key)).toEqual(
-        expect.stringMatching(/claim-status:scenarios.scenario[0-9]+.description/),
-      )
+      expect(getClaimStatusHeading(key)).toEqual(expect.stringMatching(/claim-status:scenarios.scenario[0-9]+.heading/))
     }
   })
 })
+
+// Test getClaimStatusSummary()
+// Note: These tests test against real content
```
rocketnova, line 15: @lomky I'm feeling a little 😬 about these tests that require real content (i.e. `public/locales/en/claim-status.json`, but not sure what the better solution is since the code explicitly relies on that file. I did write this [alternative set of tests](https://github.com/cagov/ui-claim-tracker/pull/314/commits/de46afe6951b358681783ad32a7ccecf480fa8c4). Let me know if you think this is better (and then help me resolve the typescript errors/warnings in them...)

utils/getClaimStatus.tsx
```diff
@@ -2,22 +2,65 @@
...
+export function getClaimStatusSummary(scenarioType: ScenarioType): TransLineProps {
+  const transLineProps: TransLineProps = {
+    i18nKey: getTranslationPrefix(scenarioType) + '.summary.text',
+  }
+  // Explicitly initialize to an empty array.
+  transLineProps.links = []
+
+  // Explicitly cast the scenarioType into one of the keys of claim-status.json
+  // (i.e. scenario1 | scenario2 etc)
+  const scenarioString = ScenarioType[scenarioType].toLowerCase() as keyof typeof claimStatusJson.scenarios
```
rocketnova, line 38: Also, if you are like me and ever struggled with `keyof typeof`, here is a very good [stackoverflow  on it](https://stackoverflow.com/a/62764510).

types/common.tsx
```diff
@@ -1,7 +1,19 @@
...
+export interface TransLineProps {
+  i18nKey: I18nString
+  links?: string[] | null
+}
+
+// Types for translation file JSON
+export interface TextOptionalLink {
+  text: string
+  links?: string[]
+}
```
rocketnova, line 14: This name is pretty terrible: `TextOptionalLink`. Please feel free to suggest alternatives!

components/ClaimStatus.tsx
lomky: perfect

utils/getUrl.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import urls from '../public/urls.json'
+
+// Type alias for the keys in urls.json
+export type UrlType = keyof typeof urls
+
+/**
+ * Get url from urls.json
+ */
+export default function getUrl(key: UrlType): string | undefined {
```
kalvinwang, line 13: I don't know how we want to handle this, and we could wait to refactor until it we have the actual different URLs, but I believe a bunch of EDD URLs are different for English vs Spanish (like not a query parameter, just a different URL for the same page)

kalvinwang (commented):
Just adding a few minor initial comments!

utils/getClaimStatus.tsx
lomky: I feel like the casting isn't the right way to do this, digging in a bit

lomky: nope, looks like this is right, but it does seem fragile. This feels like a probably anyone consuming an API would run into, I wonder what they've done?

utils/getUrl.tsx
lomky: those URLs can live in the locale files!

lomky (approved):
No blockers, but some thoughts!

components/ClaimStatus.tsx
```diff
@@ -2,20 +2,24 @@ import { useTranslation } from 'next-i18next'
 
 import { NextSteps } from './NextSteps'
 import { TextLine } from './TextLine'
+import { TransLine } from './TransLine'
```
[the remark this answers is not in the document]
lomky: Does that add clarity or just introduce HTML as an alias for the Trans component, since this is just a wrapper function?

tests/utils/apiGatewayStub.test.tsx
```diff
@@ -8,32 +8,27 @@ describe('The API gateway stub response for the Pending Determination scenario',
...
-  it('is correct for Scenario 7', () => {
-    const response = apiGatewayStub(ScenarioType.Scenario7)
+describe('The API gateway stub response for the Generic Pending state', () => {
+  it('is correct for Scenario 4', () => {
+    const response = apiGatewayStub(ScenarioType.Scenario4)
     expect([null, [], false, undefined]).toContain(response.pendingDetermination)
-    expect(response.hasPendingWeeks).toBe(false)
-    expect(response.hasCertificationWeeksAvailable).toBe(false)
+    expect(response.hasPendingWeeks).toBe(true)
+    expect([true, false]).toContain(response.hasCertificationWeeksAvailable)
```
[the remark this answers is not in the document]
rocketnova: Yes and it cannot be `null`

tests/utils/getClaimStatus.test.tsx
rocketnova: > is the problem here that we're ultimately testing the middle part of what's ultimately a visual assertion? Like what do these tests give us that the snapshot tests won't also catch?

🤔 Good point. Since we're doing snapshot tests for the `<ClaimStatus>` component for all the scenarios, I think these tests aren't actually that useful.

types/common.tsx
```diff
@@ -1,7 +1,19 @@
 // Type aliases
 export type I18nString = string
 
-// Type interfaces for API gateway result
+// Types for TransLine component
+export interface TransLineProps {
+  i18nKey: I18nString
+  links?: string[] | null
```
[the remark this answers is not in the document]
rocketnova: I think it was because I thought the `links` argument might get passed `null`. I'm currently initializing it in `TransLine.tsx` as null, but I think I can initialize it as `[]` instead.

types/common.tsx
rocketnova: `LinkableText`?

lomky: `HereThereBeLinkMaybe`? I don't feel like I have a good one 🤷🏼

rocketnova: I'm going to stick with `TextOptionalLink` for this PR. Let's talk synchronously about it if anyone feels strongly about renaming this.

utils/getClaimStatus.tsx
rocketnova: In the next PR, I've found a way to reduce the number of `as` casting to only cases where I'm trying to refer to using variables, so I think it's better.

components/ClaimStatus.tsx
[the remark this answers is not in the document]
kalvinwang: Ah I meant that it would be clearer to have Html as an alias for Trans, since HtmlLine vs TextLine is intuitively understandable (one is for pure text, the other when there's html) whereas HtmlLine vs TransLine is not unless you already know the i18next nomenclature.

Issue #317, opened 2021-07-15
Rename backend functions


rocketnova:
### Description

Let's rename them `buildTheThing` rather than `getTheThing`. See discussion https://github.com/cagov/ui-claim-tracker/pull/307#discussion_r669850557

### Acceptance Criteria

- [ ] Use `getTheThing` for calling someone and `buildTheThing` when constructing something ourselves

lomky:
I think we've moved past this - closing as wontfix

PR #318, opened 2021-07-16
Implement next steps logic

rocketnova:
## Ticket

Resolves #314 

## Changes

- Adds "Your Next Steps" and "EDD Next Steps" to Claim Status
- Adds support for `<em>` elements in `<Trans>` components
- Updates `<NextSteps>` component to expect to create text that might contain links
- Updates the Typescript target to `es6`

## Context

This PR adds in support for the latest next steps content: "Your Next Steps" and "EDD Next Steps". 

A major change to note is that I've updated the Typescript target from `es5` to `es2015` in order to support [iterating over an `IterableIterator` type](https://stackoverflow.com/questions/49218765/typescript-and-iterator-type-iterableiteratort-is-not-an-array-type/49221814). Let me know if this is going to be an issue for us.

Note: This PR **does not** include the conditional next step about certification (coming in a future PR).

## Testing

`yarn test`

PLUS

While #299 is not resolved, comment out these lines, then run `yarn storybook`, and check that each scenario displays the next steps from the latest content (including correct links): https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98


components/ClaimStatus.tsx
```diff
@@ -23,10 +29,18 @@ export const ClaimStatus: React.FC<ClaimStatusProps> = ({ loading = false, headi
...
-          <NextSteps loading={loading} nextSteps={nextSteps} />
+          <NextSteps loading={loading} nextSteps={yourNextSteps} />
+        </div>
+      </div>
+      <div className="status-box">
+        <div className="topbar">
+          <h3 className="next-steps">{t('claim-status.edd-next-steps')}</h3>
+        </div>
+        <div className="explanation">
+          <NextSteps loading={loading} nextSteps={eddNextSteps} />
```
rocketnova, line 43: @lomky Again, I just tossed all of this in here and assumed you'd do some finessing.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
rocketnova, line 3: Let me know if this is going to be an issue!

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
 import getUrl, { UrlType } from './getUrl'
 
+type StepType = 'your-next-steps' | 'edd-next-steps'
+
+interface ClaimStatusScenarioJson {
+  heading: string
+  summary: TextOptionalLink
+  'your-next-steps': TextOptionalLink[]
+  'edd-next-steps': TextOptionalLink[]
+}
```
rocketnova, line 17: I introduced some stronger types here so I was able to both avoid the very loosely typed [recursion approach](https://github.com/cagov/ui-claim-tracker/blob/18610707d4d7a6e6367b0e45f079f6a06cfba2ca/utils/getClaimStatus.tsx) (check out `getChildJson()`) I was trying out and end up with slightly tighter typing than in PR #314.

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
+    steps.push(buildTransLineProps(value, buildI18nKey(keys)))
   }
-  return transLineProps
+  return steps
 }
 
 /**
  * Get combined Claim Status content.
  */
 export default function getClaimStatus(scenarioType: ScenarioType): ClaimStatusContent {
```
rocketnova, line 108: @lomky I've renamed most of the functions in this file to `buildTheThing` but I left the default function for now to avoid renaming files and imports. We can do more renaming in #317

components/ClaimStatus.tsx
lomky: oh this will be a fun merge in! 😅

lomky: quick guidance since I refactored a component under ya:

A lot of the structure here got nixed:
 - the `h3, `status-box`, `explanation` & `topbar` gone

now you just call `NextSteps` and pass it the next steps and header

So i _think_ this component return should end up just looking like:

```tsx
    <div className="claim-status claim-section">                                                                                        
      <h2>{t('claim-status.title')}</h2>                                                                                                
      <div className="pending-status claim-subsection">                                                                                 
        <TextLine loading={loading} header text={t(heading)} />                                                                         
      </div>                                                                                                                            
      <div className="summary">                                                                                                         
        <TransLine loading={loading} i18nKey={summary.i18nKey} links={summary.links} />                                                 
      </div>                                                                                                                            
      <NextSteps loading={loading} header={t('claim-status.your-next-steps')} nextSteps={yourNextSteps} />                                       
      <NextSteps loading={loading} header={t('claim-status.edd-next-steps')} nextSteps={eddNextSteps} />                                       
    </div>  
```

components/NextSteps.tsx
```diff
@@ -1,11 +1,13 @@
 import { Shimmer } from './Shimmer'
```
lomky, line 1: ooo another sticky merge: I think you take priority for on the nextStep typing, but you should pull in the structure as redone on main (sorry!)

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
+
+  const scenarioObject = claimStatusJson.scenarios[scenarioString]
 
-  return statusContent
+  return {
+    heading: buildClaimStatusHeading(scenarioType),
+    summary: buildClaimStatusSummary(scenarioObject, scenarioString),
+    yourNextSteps: buildNextSteps(scenarioObject, scenarioString, 'your-next-steps'),
+    eddNextSteps: buildNextSteps(scenarioObject, scenarioString, 'edd-next-steps'),
+  }
```
lomky, line 120: github mangled this file's diff so much that I had an excuse to use `vimdiff` for the first time in _ages_! 

Really like this `return` refactor. 👨🏼‍🍳👌🏼

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
- * Note: This function has a lot of verbosity due to Typescript + eslint.
+ * Build a list of urls from keys in the json translation files.
  */
-export function getClaimStatusSummary(scenarioType: ScenarioType): TransLineProps {
-  const transLineProps: TransLineProps = {
-    i18nKey: getTranslationPrefix(scenarioType) + '.summary.text',
+export function buildTransLineLinks(linkKeys: string[] | undefined): string[] {
+  const links: string[] = []
+  if (linkKeys) {
+    if (linkKeys && linkKeys.length > 0) {
```
lomky, line 46: Do we need this existence check twice?

tsconfig.json
lomky: I think we should probably chat through the browser support implications at eng sync, but I don't inherently disagree!

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
+      for (const linkKey of linkKeys) {
+        // Explicitly cast to one of the allowed keys in urls.json
+        const key = linkKey as UrlType
```
kalvinwang, line 49: What do you think about doing this casting inside of `getClaimStatus` instead of waiting until here? The goal being to consolidate the need-to-do-this-for-typescript code. Not just the casting I guess, really this entire function.

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
  */
-export function getClaimStatusSummary(scenarioType: ScenarioType): TransLineProps {
-  const transLineProps: TransLineProps = {
-    i18nKey: getTranslationPrefix(scenarioType) + '.summary.text',
+export function buildTransLineLinks(linkKeys: string[] | undefined): string[] {
+  const links: string[] = []
+  if (linkKeys) {
+    if (linkKeys && linkKeys.length > 0) {
+      for (const linkKey of linkKeys) {
+        // Explicitly cast to one of the allowed keys in urls.json
```
kalvinwang, line 48: What do you think about doing this casting inside of `getClaimStatus` instead of waiting until here? The goal being to consolidate the need-to-do-this-for-typescript code. I'd actually like to expand this comment with `; this is necessary for Typescript due to the variable in claimStatusJson.scenarios[scenarioString]`, but that  context would make more sense if this were moved. Just a thought

Not just the casting I guess, really this entire function.

kalvinwang (approved):
lgtm, left a bunch of comments!

public/locales/es/claim-status.json
```diff
@@ -4,25 +4,83 @@
...
     },
     "scenario6": {
       "heading": "Semanas disponibles para certificar",
       "summary": {
         "text": "Tiene semanas disponibles para <0> certificar los beneficios </0>."
-      }
+      },
+      "your-next-steps": [
+        {
+          "text": "Seleccione <strong> Certificar para obtener beneficios </strong> en la sección de notificaciones de la página de inicio de UI Online."
```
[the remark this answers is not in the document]
rocketnova: Good catch. I'll fix these.

utils/getClaimStatus.tsx
kalvinwang: I am not sure this is an example of it, but reading "[push the burden of proof upward as far as possible](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), but no further" reminded me of this

utils/getClaimStatus.tsx
rocketnova: Whoops that was a mistake. I'll fix.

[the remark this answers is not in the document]
rocketnova: Good catch. Line 46 was a mistake. I'll correct.

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
 /**
- * Get Claim Status summary.
- *
- * Note: This function has a lot of verbosity due to Typescript + eslint.
+ * Build a list of urls from keys in the json translation files.
  */
-export function getClaimStatusSummary(scenarioType: ScenarioType): TransLineProps {
-  const transLineProps: TransLineProps = {
-    i18nKey: getTranslationPrefix(scenarioType) + '.summary.text',
+export function buildTransLineLinks(linkKeys: string[] | undefined): string[] {
```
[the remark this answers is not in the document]
rocketnova: I wanted to allow `linkKeys` to be `undefined` so that we don't have to populate the translation files with a bunch of lines that look like this:

```
"your-next-steps": [
  {
    "text": "Text here",
    "links": [] <--- noisy empty arrays
  },
  ...
],
```

utils/getClaimStatus.tsx
```diff
@@ -7,66 +7,115 @@ import { ClaimStatusContent, I18nString, TextOptionalLink, TransLineProps } from
...
+  const links: string[] = []
+  if (linkKeys) {
+    if (linkKeys && linkKeys.length > 0) {
+      for (const linkKey of linkKeys) {
+        // Explicitly cast to one of the allowed keys in urls.json
+        const key = linkKey as UrlType
+        const url = getUrl(key)
+        if (url) {
```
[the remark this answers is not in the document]
rocketnova: Well, it shouldn't happen, but it might very well happen if someone makes a typo in `claim-status.json`. I don't think we should throw an error because I think the end-user should still see all the rest of the content, even if one of the links is missing, but I do think we should probably log it and get some kind of alert or something. I'll add some logging to it.

utils/getClaimStatus.tsx
rocketnova: Hmmm because this function is really about url keys rather than the `variable in claimStatusJson.scenarios[scenarioString]`, I could see moving the cast over to `getUrl.tsx` but not into `getClaimStatus()`. I'll do that.

PR #325, opened 2021-07-20
Moves Header and Footer links to url file, Removes unneeded locale

lomky:
- Gets bespoke and redundant URLs out of the components and into the public urls file. 
- Moves hacky urls-in-locale to the urls file
- Removes the locale querystring from the UIO Home link
- Renames the header test to match naming convention

Testing: snapshots changes are limited to locale removal - otherwise all URLs are unchanged


public/urls.json
```diff
@@ -1,8 +1,18 @@
 {
```
lomky, line 1: I couldn't resist sorting this. please lemme know if I shouldn't have

public/urls.json
```diff
@@ -1,8 +1,18 @@
...
-  "edd-ui-claim-status": "https://www.edd.ca.gov/unemployment/claim-status.htm",
-  "edd-ui-eligibility": "https://edd.ca.gov/Unemployment/Eligibility.htm",
+  "ca-gov": "https://www.ca.gov/",
+  "edd-about-accessibility": "https://www.edd.ca.gov/About_EDD/Accessibility.htm",
+  "edd-about-conditions-of-use": "https://www.edd.ca.gov/About_EDD/Conditions_Of_Use.htm",
+  "edd-about-contact": "https://www.edd.ca.gov/About_EDD/Contact_EDD.htm",
+  "edd-about-privacy-policy": "https://www.edd.ca.gov/About_EDD/Privacy_Policy.htm",
+  "edd-ca-gov": "https://edd.ca.gov",
   "edd-covid": "https://www.edd.ca.gov/about_edd/coronavirus-2019/benefit-year-end.htm",
```
lomky, line 8: following the convention of my other links to `https://www.edd.ca.gov/About_EDD/[...]` this would get renamed to `edd-about-covid` but I didn't want to step on any toes

rocketnova: Feel free to rename!

public/urls.json
rocketnova: No worries. Thanks for sorting!

components/Footer.tsx
```diff
@@ -12,28 +13,16 @@ export const Footer: React.FC = () => {
...
-            <Nav.Link target="_blank" rel="noopener noreferrer" href="https://www.edd.ca.gov/About_EDD/Contact_EDD.htm">
+            <Nav.Link target="_blank" rel="noopener noreferrer" href={getUrl('edd-about-contact')}>
               {t('footer.contact')}
             </Nav.Link>
-            <Nav.Link
-              target="_blank"
-              rel="noopener noreferrer"
-              href="https://www.edd.ca.gov/About_EDD/Conditions_Of_Use.htm"
-            >
+            <Nav.Link target="_blank" rel="noopener noreferrer" href={getUrl('edd-about-conditions-of-use')}>
```
rocketnova, line 19: I see that we have this pattern a lot:

```
<Nav.Link target="_blank" rel="noopener noreferrer" href={something}>
```

Is it worth breaking into its own mini component?

rocketnova (approved):
LGTM! One suggestion. Take it or leave it!

public/urls.json
```diff
@@ -1,8 +1,18 @@
...
-  "edd-ui-certify": "https://edd.ca.gov/Unemployment/certify.htm",
   "edd-faq-bpo": "https://edd.ca.gov/FAQ_-_Benefit_Programs_Online.htm",
-  "edd-ui-reopen": "https://edd.ca.gov/Unemployment/Reopen-A-Claim.htm"
+  "edd-help-new-claim": "https://uio.edd.ca.gov/UIO/Pages/Public/help/index.htm#t=en-US/Public/NewClaim/UIOnlineNewClaimLandingPage.htm",
```
rocketnova, line 10: @lomky Just thought of this: Do you think this URL might need to be different depending on mobile vs desktop?

lomky, line 10: Should this have a mobile v desktop?

public/urls.json
lomky:
```suggestion
  "edd-about-covid": "https://www.edd.ca.gov/about_edd/coronavirus-2019/benefit-year-end.htm",
```

public/urls.json
lomky: I have no way to check - you can't access UIMobile unless you have a real active claim, so I'm not sure if they even have this header

components/Footer.tsx
lomky: maybe? if we use it going forward I'm game but I don't feel like it warrants a refactor to do so!

PR #330, opened 2021-07-21
Add ability to have language-dependent urls

### LINKED ISSUE #324: Link to different pages depending on the language
### Description

The current url lookup makes an assumption that the URLs to link to are the same for all languages. This is not true. We need to display different urls for English and for Spanish.

### Acceptance Criteria

- [ ] The correct url is linked to for each language

rocketnova:
## Ticket

Resolves #324 

## Changes

- Move language-dependent urls into `common.json`
- Move url resolution logic from `getClaimStatus` to `<TransLine>`
- Update scenario content to latest content from UIB
- Move logic to determine whether the user came from UIO Mobile from the `<Home>` component to `getServerSideProps()`
- Add `userArrivedFromUioMobile` argument to many components
- Split `TransLineContent` type from `TransLineProps` type to better mirror other types in `common`
- Add `userArrivedFromUioMobile` boolean control to the Page story

## Context

This PR has a lot of changes, but they can be loosely organized into three buckets:

1. Changes to support language-dependent url resolution
2. Changes to support mobile vs non-mobile url resolution
3. Renaming types

## Testing

`yarn test`

PLUS

While #299 is not resolved, comment out the following lines, then run `yarn storybook`, and verify that the links to UIO mobile work when the toggle is set to `true`: https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98

PLUS

1. Run `yarn dev`
2. Load localhost:3000
3. Click "En español"
4. Make sure the links change to their `-espanol` version

### DIFF
--- a/components/ClaimSection.tsx
+++ b/components/ClaimSection.tsx
@@ -4,13 +4,20 @@ import { ScenarioContent } from '../types/common'
 
 export interface ClaimSectionProps extends ScenarioContent {
   loading: boolean
+  userArrivedFromUioMobile: boolean
 }
 
-export const ClaimSection: React.FC<ClaimSectionProps> = ({ loading = false, statusContent, detailsContent }) => {
+export const ClaimSection: React.FC<ClaimSectionProps> = ({
+  loading = false,
+  userArrivedFromUioMobile = false,
+  statusContent,
+  detailsContent,
+}) => {
   return (
     <div className="claim-section">
       <ClaimStatus
         loading={loading}
+        userArrivedFromUioMobile={userArrivedFromUioMobile}
         heading={statusContent.heading}
         summary={statusContent.summary}
         yourNextSteps={statusContent.yourNextSteps}
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -1,7 +1,6 @@
 import { Trans, useTranslation } from 'react-i18next'
 import React from 'react'
 import { Shimmer } from './Shimmer'
-import { useRouter } from 'next/router'
 import { TransLineContent } from '../types/common'
 import getUrl from '../utils/getUrl'
 
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -1,12 +1,31 @@
 import { Trans, useTranslation } from 'react-i18next'
 import React from 'react'
 import { Shimmer } from './Shimmer'
-import { TransLineContent } from '../types/common'
+import { I18nString, TransLineContent } from '../types/common'
 import getUrl from '../utils/getUrl'
 
 export interface TransLineProps extends TransLineContent {
   loading: boolean
   userArrivedFromUioMobile: boolean
+}
+
+/**
+ * Handle url resolution.
+ */
+function resolveUrl(link: I18nString, userArrivedFromUioMobile: boolean) {
+  // Special case for UIO homepage links.
+  if (link === 'uio-home') {
+    const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
+    if (uioHomeLink) {
+      // If the link is for UIO homepage, do a direct getUrl() lookup.
+      // Do not pass the looked up url through t() because t() will mangle the url.
+      return uioHomeLink
+    }
+  }
+
+  // Otherwise, use t() to lookup the correct language-dependent url.
+  const { t } = useTranslation(['common', 'claim-details', 'claim-status'])
+  return t(link)
 }
 
 export const TransLine: React.FC<TransLineProps> = ({
@@ -19,25 +38,10 @@ export const TransLine: React.FC<TransLineProps> = ({
     return <Shimmer width={120} height={15} baseColor="#B6B2B2" shimColor="#656565" borderRadius={3} />
   }
 
-  const { t } = useTranslation(['common', 'claim-details', 'claim-status'])
-
   const linkComponents: JSX.Element[] = []
   if (links && links.length > 0) {
     for (const link of links) {
-      // Special case for UIO homepage links.
-      let href = ''
-      if (link === 'uio-home') {
-        const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
-        if (uioHomeLink) {
-          // If the link is for UIO homepage, do a direct getUrl() lookup.
-          // Do not pass the looked up url through t() because t() will mangle the url.
-          href = uioHomeLink
-        }
-      }
-      // Otherwise, use t() to lookup the correct language-dependent url.
-      else {
-        href = t(link)
-      }
+      const href = resolveUrl(link, userArrivedFromUioMobile)
       // Disabling some linting rules for this line. The anchor <a> element will
       // be interpolated by <Trans>.
       /* eslint-disable jsx-a11y/anchor-has-content */
--- a/components/TransLine.tsx
+++ b/components/TransLine.tsx
@@ -1,4 +1,4 @@
-import { Trans } from 'react-i18next'
+import { Trans, useTranslation } from 'react-i18next'
 import React from 'react'
 import { Shimmer } from './Shimmer'
 import { TransLineProps } from '../types/common'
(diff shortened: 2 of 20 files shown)


components/TransLine.tsx
```diff
@@ -1,22 +1,50 @@
...
-    // be interporlated by <Trans>.
-    /* eslint-disable jsx-a11y/anchor-has-content */
-    /* eslint-disable react/self-closing-comp */
-    linkComponents = links.map((link) => <a href={link} key={link}></a>)
-    /* eslint-enable jsx-a11y/anchor-has-content */
-    /* eslint-enable react/self-closing-comp */
+    for (const link of links) {
+      // Special case for UIO homepage links.
+      let href = ''
+      if (link === 'uio-home') {
```
rocketnova, line 29: @lomky I'm totally open to changing this string/approach if you want. Since right now it seems like there is only one UIO url that changes based on mobile vs desktop, I made it a single special case and assumed we would refactor if/when we get more urls.

components/TransLine.tsx
```diff
@@ -1,22 +1,50 @@
...
+        const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
+        if (uioHomeLink) {
+          // If the link is for UIO homepage, do a direct getUrl() lookup.
+          href = uioHomeLink
```
rocketnova, line 33: Note: It's important not to pass this url through `t()` as the url gets mangled.

public/urls.json
```diff
@@ -5,14 +5,9 @@
...
   "edd-help-new-claim": "https://uio.edd.ca.gov/UIO/Pages/Public/help/index.htm#t=en-US/Public/NewClaim/UIOnlineNewClaimLandingPage.htm",
   "edd-log-in": "https://portal.edd.ca.gov/WebApp/Login?resource_url=",
   "edd-log-out": "https://portal.edd.ca.gov/WebApp/Logout",
-  "edd-ui-certify": "https://edd.ca.gov/Unemployment/certify.htm",
-  "edd-ui-claim-status": "https://www.edd.ca.gov/unemployment/claim-status.htm",
-  "edd-ui-eligibility": "https://edd.ca.gov/Unemployment/Eligibility.htm",
-  "edd-ui-reopen": "https://edd.ca.gov/Unemployment/Reopen-A-Claim.htm",
   "uio-home-url-desktop": "https://uio.edd.ca.gov/UIO/Pages/Public/ExternalUser/UIOnlineLandingPage.aspx",
   "uio-home-url-mobile": "https://uiom.edd.ca.gov/UIOM/Pages/Public/ExternalUser/UIOMobileLandingPage.aspx"
 }
```
rocketnova, line 13: @lomky I didn't end up restructuring `urls.json` in this PR because I was already making changes in so many files. I'm totally open to doing that in a future PR.

tests/components/ClaimStatus.test.tsx
```diff
@@ -18,39 +19,63 @@ function renderClaimStatusComponent(statusContent: ClaimStatusContent): string {
...
 
-function testClaimStatus(scenarioType: ScenarioType, hasCertificationWeeksAvailable: boolean): string {
+function testClaimStatus(
+  scenarioType: ScenarioType,
+  hasCertificationWeeksAvailable: boolean,
+  userArrivedFromUioMobile: boolean,
+): string {
   const scenarioContent = getScenarioContent(apiGatewayStub(scenarioType, hasCertificationWeeksAvailable))
-  return renderClaimStatusComponent(scenarioContent.statusContent)
+  return renderClaimStatusComponent(scenarioContent.statusContent, userArrivedFromUioMobile)
```
rocketnova, line 28: Note: I've added a lot of snapshots here, but I thought it didn't hurt to have extra/added coverage.

components/TransLine.tsx
kalvinwang: would it be useful to leave this as an inline comment for future readers? `...getUrl() lookup because t() mangles the url`

pages/index.tsx
```diff
@@ -97,6 +94,7 @@ export const getServerSideProps: GetServerSideProps = async ({ req, locale }) =>
   }
 
   logger.info(req)
+  logger.info(query)
```
kalvinwang, line 97: No idea what this query object looks like other than it contains `from`, but would it make sense to log it in the same call as `req` above, if that's possible? Just thinking it would be easier to read logs with combined request/queries...

components/TransLine.tsx
```diff
@@ -1,22 +1,50 @@
...
+        }
+      }
+      // Otherwise, use t() to lookup the correct language-dependent url.
+      else {
+        href = t(link)
+      }
+      // Disabling some linting rules for this line. The anchor <a> element will
+      // be interporlated by <Trans>.
```
kalvinwang, line 41: thanks for the comment! small typo, `interpolated`

utils/getClaimStatus.tsx
```diff
@@ -37,32 +36,19 @@ export function buildClaimStatusHeading(scenarioType: ScenarioType): I18nString
...
  */
-export function buildTransLineProps(json: TextOptionalLink, i18nKey: I18nString): TransLineProps {
-  return {
+export function buildTransLineContent(json: TextOptionalLink, i18nKey: I18nString): TransLineContent {
+  const props: TransLineContent = {
     i18nKey: i18nKey,
-    links: buildTransLineLinks(json.links),
   }
+
+  if (json.links) {
```
kalvinwang, line 47: so much cleaner [in this file], nice!

kalvinwang (approved):
lgtm, thank you!

public/urls.json
lomky: whenever it comes up - I'm in no rush!

components/ClaimSection.tsx
```diff
@@ -4,13 +4,20 @@ import { ScenarioContent } from '../types/common'
 
 export interface ClaimSectionProps extends ScenarioContent {
   loading: boolean
+  userArrivedFromUioMobile: boolean
 }
 
-export const ClaimSection: React.FC<ClaimSectionProps> = ({ loading = false, statusContent, detailsContent }) => {
+export const ClaimSection: React.FC<ClaimSectionProps> = ({
+  loading = false,
+  userArrivedFromUioMobile = false,
```
lomky, line 12: I FINALLY FOUND IT!

I knew there must be a way to avoid this pass-pass-pass through the components:

> Context is designed to share data that can be considered “global” for a tree of React components, such as the current authenticated user, theme, or preferred language. For example, in the code below we manually thread through a “theme” prop in order to style the Button component

[React Context](https://reactjs.org/docs/context.html)

lomky: to be clear: if we do this I suggest we do this as a follow on, not in this PR

components/TransLine.tsx
lomky: I think its fine since this is the one exception, but I might pull it out into a function to keep this function readable?

pages/index.tsx
```diff
@@ -115,12 +113,17 @@ export const getServerSideProps: GetServerSideProps = async ({ req, locale }) =>
     errorCode = 500
   }
 
+  // Note whether the user came from the main UIO website or UIO Mobile, and match
+  // that in our links back out to UIO.
+  const userArrivedFromUioMobile = query?.from === 'uiom'
```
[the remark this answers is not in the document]
lomky: oh this is way better!

tests/components/ClaimStatus.test.tsx
```diff
@@ -18,39 +19,63 @@ function renderClaimStatusComponent(statusContent: ClaimStatusContent): string {
...
 
 describe('Scenario 1', () => {
-  it('matches when there are weeks to certify', () => {
+  it('matches when there are weeks to certify, on desktop', () => {
     expect(testClaimStatus(ScenarioType.Scenario1, true)).toMatchSnapshot()
   })
-  it("matches when there aren't weeks to certify", () => {
+  it('matches when there are weeks to certify, on mobile', () => {
+    expect(testClaimStatus(ScenarioType.Scenario1, true, true)).toMatchSnapshot()
+  })
```
lomky, line 37: ooo love it

pages/index.tsx
rocketnova: This is surprisingly not straight forward. I've opened a separate ticket for it https://github.com/cagov/ui-claim-tracker/issues/332

Issue #342, opened 2021-07-27
Format and display Next Steps with nested bullets


rocketnova:
### Description

This ticket is split from #252

### Acceptance Criteria

- [ ] Business logic correctly provides the nested bullet content to the FE components

Issue #341, opened 2021-07-27
Format and display appointment date and time for Scenario 2


rocketnova:
### Description

This ticket splits out from #252.

### Acceptance Criteria

- [ ] Format and display the appointment date and time in Scenario 2
- [ ] The date should be displayed in the proper locale

Issue #329, opened 2021-07-20
Integrate Final Content into the Claim Tracker Application


lomky:
### Description

As we are finalizing the content for the initial release, work with Content to get the final strings into the application

### Acceptance Criteria

- [ ] Final English strings are in the application
  - [ ] Common strings
  - [ ] scenario specific strings
- [ ] Translated strings are in the application

kalvinwang:
@rocketnova oops didn't see you just self-assigned it! anyway i think i'm updating it accurately

PR #343, opened 2021-07-27
Implement pending determination scenario logic

### LINKED ISSUE #252: Implement pending determination phases
### Description

This ticket splits out the conditional logic for the pending determination phases from #242 and should display the correct content for a given claimant, and add the content (untranslated).

### Acceptance Criteria

- [ ] Conditional logic for pending determination phases implemented

rocketnova:
## Ticket

Resolves #252 

## Changes

- Add business logic for Scenarios 1–3, including:
  - Identifying whether the Determination Status is pending
  - Comparing dates and times between `pendingDetermination` objects
  - Handling multiple `pendingDetermination` objects
- Add [`date-fns-tz`](https://github.com/marnusw/date-fns-tz) as a dependency to handle time zones
- Add date parsing and formatting utility (moved out of `getClaimDetails.tsx`)
- Add time slot utility
- Add support for Claim Status summary to have multiple paragraphs
- Add latest English language scenario content
- Update `apiGatewayStub` to support Scenarios 1–3
- Add [`mockdate`](https://github.com/boblauer/MockDate) package for more robust mocking of `Date.now()`

## Context

This PR adds a lot of functionality to support the 3 pending determination/determination interview scenarios. The logic is bit complicated, so I'm happy to walk you through it. The highlights are:

- If there are multiple `pendingDetermination` objects, Scenario 2 takes priority, then Scenario 3, then Scenario 1.
- If there are multiple objects that fit Scenario 2, show the one that has the soonest appointment.
- It's possible for there to be invalid `pendingDetermination` objects, in which case, fall through to the logic for Scenarios 4–6.

The content turned out to actually be of a slightly different format from the Generic Pending and Base State scenarios, so I've split out 2 additional tickets:
- #341 
- #342 

## Testing

`yarn test`

PLUS

While #299 is not resolved, comment out the following lines, then run `yarn storybook`, and verify that the links the 3 pending determination scenarios display the latest content (with the exceptions of the upcoming appointment date and the nested bullets): https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98

### DIFF
--- a/components/ClaimStatus.tsx
+++ b/components/ClaimStatus.tsx
@@ -27,12 +27,16 @@ export const ClaimStatus: React.FC<ClaimStatusProps> = ({
         <TextLine loading={loading} header text={t(heading)} />
       </div>
       <div className="summary">
-        <TransLine
-          loading={loading}
-          userArrivedFromUioMobile={userArrivedFromUioMobile}
-          i18nKey={summary.i18nKey}
-          links={summary.links}
-        />
+        {summary.map((paragraph, index) => (
+          <div key={index} className="">
+            <TransLine
+              loading={loading}
+              userArrivedFromUioMobile={userArrivedFromUioMobile}
+              i18nKey={paragraph.i18nKey}
+              links={paragraph.links}
+            />
+          </div>
+        ))}
       </div>
       <NextSteps
         loading={loading}
--- a/public/locales/en/claim-status.json
+++ b/public/locales/en/claim-status.json
@@ -2,9 +2,11 @@
   "scenarios": {
     "scenario1": {
       "heading": "Pending Eligibility — Phone Interview Will Be Scheduled",
-      "summary": {
-        "text": "We identified an issue that could make you ineligible for benefits. We need to confirm your eligibility in a phone interview."
-      },
+      "summary": [
+        {
+          "text": "We identified an issue that could make you ineligible for benefits. We need to confirm your eligibility in a phone interview."
+        }
+      ],
       "your-next-steps": [
         {
           "text": "Confirm we have your current phone number and mailing address. Go to the <0>UI Online homepage</0>, select <strong>Profile</strong> from the main menu, then select <strong>Contact Information</strong>.",
@@ -20,11 +22,82 @@
         }
       ]
     },
+    "scenario2": {
+      "heading": "Pending Eligibility — Phone Interview Scheduled",
+      "summary": [
+        {
+          "text": "We identified a potential issue that could make you ineligible for benefits. You have a phone interview scheduled for:"
+        },
+        {
+          "text": "<strong>Important</strong>: If you miss your phone interview, a decision will be made based on the available facts, which could result in your unemployment benefits being delayed or denied."
+        }
+      ],
+      "your-next-steps": [
+        {
+          "text": "Confirm we have your current phone number. Go to the <0>UI Online homepage</0>, select <strong>Profile</strong> from the main menu, then select <strong>Contact Information</strong>.",
+          "links": ["uio-home"]
+        },
+        {
+          "text": "Prepare for the interview. The <em>Notification of Unemployment Insurance Benefits Eligibility Interview</em> (DE 4800) includes the questions the interviewer is most likely to ask."
+        },
+        {
+          "text": "If you need to reschedule your interview, select <strong>Reschedule</strong> in the Appointments section on your <0>UI Online homepage</0>. You must reschedule at least one day before the interview.",
+          "links": ["uio-home"]
+        }
+      ],
+      "edd-next-steps": [
+        {
+          "text": "We will call you during your scheduled interview time. Your caller ID may show “St of CA EDD” or the UI Customer Service number 1-800-300-5616.",
+          "sub-bullets": [
+            {
+              "text": "If you do not receive a call from the EDD at your scheduled appointment time, we may have canceled your appointment because we confirmed your eligibility or resolved the issue before your interview."
+            },
+            {
+              "text": "If your appointment has been canceled, it will no longer display on your UI Online homepage."
+            }
+          ]
+        }
+      ]
+    },
+    "scenario3": {
+      "heading": "Pending Eligibility — Under Review",
+      "summary": [
+        {
+          "text": "Your phone interview time has passed and your eligibility for benefits is under review."
+        },
+        {
+          "text": "<strong>Important</strong>: If you missed your phone interview, a decision will be made based on the available facts, which could result in your unemployment benefits being delayed or denied."
+        }
+      ],
+      "your-next-steps": [
+        {
+          "text": "Allow up to 10 days for the EDD to make a decision."
+        }
+      ],
+      "edd-next-steps": [
+        {
+          "text": "We will determine your eligibility.",
+          "sub-bullets": [
+            {
+              "text": "If you are found eligible and no other issues are identified, we will pay you for all pending weeks."
+            },
+            {
+              "text": "If you are found not eligible, we will mail you a <em>Notice of Determination</em> (DE 1080CZ) with the reasons you were denied benefits and an <em>Appeal Form</em> (DE 1000M). If you disagree with the decision, you have the right to appeal. "
+            },
+            {
+              "text": "If you are paid benefits and are later found not eligible, we will also mail you a <em>Notice of Overpayment</em> (DE 1444) that explains why you were overpaid."
+            }
+          ]
+        }
+      ]
+    },
     "scenario4": {
       "heading": "Review Required",
-      "summary": {
-        "text": "We identified an issue and may need to determine your eligibility or verify your information."
-      },
+      "summary": [
+        {
+          "text": "We identified an issue and may need to determine your eligibility or verify your information."
+        }
+      ],
       "your-next-steps": [
         {
           "text": "Check your <0>UI Online homepage</0> and inbox for the latest updates.",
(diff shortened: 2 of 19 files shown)


components/ClaimStatus.tsx
```diff
@@ -27,12 +27,16 @@ export const ClaimStatus: React.FC<ClaimStatusProps> = ({
...
+        {summary.map((paragraph, index) => (
+          <div key={index} className="">
+            <TransLine
+              loading={loading}
+              userArrivedFromUioMobile={userArrivedFromUioMobile}
+              i18nKey={paragraph.i18nKey}
+              links={paragraph.links}
+            />
+          </div>
+        ))}
```
rocketnova, line 39: @lomky This section will need some massaging (I tossed in a `className` in an inner `<div>` for now). But perhaps worth waiting until #341 is done before working on it.

public/locales/en/claim-status.json
```diff
@@ -20,11 +22,82 @@
...
+    "scenario2": {
+      "heading": "Pending Eligibility — Phone Interview Scheduled",
+      "summary": [
+        {
+          "text": "We identified a potential issue that could make you ineligible for benefits. You have a phone interview scheduled for:"
+        },
+        {
+          "text": "<strong>Important</strong>: If you miss your phone interview, a decision will be made based on the available facts, which could result in your unemployment benefits being delayed or denied."
+        }
+      ],
```
rocketnova, line 34: Note: Summaries now are an array of `TextOptionalLink`s.

types/common.tsx
```diff
@@ -1,5 +1,6 @@
 // Type aliases
 export type I18nString = string
+export type ApiGatewayDateString = string
```
rocketnova, line 3: Note: I added this alias so I could more easily see when I was expecting a string of the format `"yyyy-MM-dd'T'HH:mm:ss"`, which is the format that the API gateway presents datestrings as.

utils/timeSlot.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import { TimeSlot } from '../types/common'
+
+/**
+ * Parse a time slot from the API gateway.
+ */
+export function parseTimeSlot(timeSlot: string): TimeSlot | null {
+  // Time slots are expected to be in the format 10-12.
+  const match = /(\d+)[-–—](\d+)/.exec(timeSlot)
```
rocketnova, line 12: Note: This line doesn't render very well in Github. The `[-–—]` is saying that either a hyphen (`-`) or an ndash (`–`) or an mdash (`—`) are acceptable.

utils/getScenarioContent.tsx
```diff
@@ -7,51 +7,186 @@
...
+
+    // It's possible for both time slots to be improperly formatted, in which case it doesn't
+    // matter which appointment is said to be first, since they are on the same date.
+    if (!isEarlier) {
+      return false
+    }
+    // Otherwise, return the appointment with the earlier time slot start time.
+    else {
+      return isEarlier
+    }
```
lomky, line 87: I think you can return `!!isEarlier` - that will return false for `false` or `null`, `true` for `true`

utils/formatDate.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Return a string that matches the API gateway format for datetimes.
+ */
+export function formatFromApiGateway(date: Date): string {
+  return format(date, apiGatewayFormat)
+}
+
+/**
+ * Create a Date object that is offset from today.
+ */
+export function getDateWithOffset(offset = 1): Date {
```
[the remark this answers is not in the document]
rocketnova: These are also used in the `apiGatewayStub` so in one of my commits, I moved them out of the test file and into this util.

public/locales/en/claim-status.json
```diff
@@ -2,9 +2,11 @@
   "scenarios": {
```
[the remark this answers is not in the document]
rocketnova: Yes! Though honestly, we can just fold it into this PR and resolve both tickets with this one.

utils/getScenarioContent.tsx
```diff
@@ -7,51 +7,186 @@
...
+ * Determine whether the Determination Status is pending.
+ *
+ * Determination Status is considered pending if:
+ * either DeterminationStatus is blank (NULL)
+ * OR DeterminationStatus is neither "Canceled" nor "Complete" nor "TRAN" nor
+ *   "INVL" nor "IDNC" nor "1277" nor "OTHR" nor "ClmCX"
+ */
+export function isDeterminationStatusPending(pendingDetermination: PendingDetermination): boolean {
+  return (
+    !pendingDetermination.determinationStatus ||
```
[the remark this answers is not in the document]
rocketnova: I _think_ I actually want to be checking for falsy rather than just blank/null. I'm not super confident in the range of values that might come back from the API gateway (see `getScenarioContent.test.tsx` for the full range of values I'm checking for). I assume this might change once we implement tighter validation with #150.

utils/formatDate.ts
```diff
@@ -0,0 +1,107 @@
...
+  const today = new Date()
+  const sometime = today.setDate(today.getDate() + offset)
+  const sometimeUtc = datetimeInUtc(sometime)
+  return sometimeUtc
```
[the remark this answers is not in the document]
rocketnova: I was thinking that it would be safer to have everything in UTC, but now that I'm actively working on the time slot parsing, I'm realizing that it might be the wrong choice. I'll revise this.

utils/getScenarioContent.tsx
rocketnova: Neat! I don't think I've seen that syntax before.

public/locales/en/claim-status.json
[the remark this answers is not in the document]
rocketnova: For the sake of time, I think I'm going to leave adding the Spanish translation for #329 as a separate PR.

tests/utils/getScenarioContent.test.tsx
```diff
@@ -1,63 +1,388 @@
...
-    const pendingDeterminationScenarioWithout = {
-      pendingDetermination: ['temporary text'],
-      hasPendingWeeks: false,
-    }
-    const scenarioTypeWithout = getScenario(pendingDeterminationScenarioWithout)
-    expect(scenarioTypeWithout).toBe(ScenarioType.Scenario1)
-  })
-
-  it('is not returned if pendingDetermination is null', () => {
+// Scenarios 1, 2, 3
```
lomky, line 52:
```suggestion
// Scenarios 1, 2, 3 negative identification tests
```

tests/utils/getScenarioContent.test.tsx
```diff
@@ -1,63 +1,388 @@
...
     for (const scenarioType of baseScenarios) {
-      expect(getScenario(apiGatewayStub(scenarioType))).toBe(scenarioType)
+      const scenarioObject = getScenario(apiGatewayStub(scenarioType))
+      expect(scenarioObject.scenarioType).toBe(scenarioType)
+    }
+  })
+})
+
+/**
+ * Test Test identifyPendingDeterminationScenario()
```
lomky, line 92:
```suggestion
 * Test identifyPendingDeterminationScenario()
```

utils/formatDate.ts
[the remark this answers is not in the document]
rocketnova: Ok after some more work, I've refactored a number of things and added _a lot more_ comments to `formatDate.ts`, reproduced here for reference.

According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date:
> A JavaScript date is fundamentally specified as the number of milliseconds that have
> elapsed since midnight on January 1, 1970, UTC. ... It's important to keep in mind
> that while the time value at the heart of a Date object is UTC, the basic methods to
> fetch the date and time or its components all work in the local (i.e. host system)
> time zone and offset.

So by default, if a date time string has no time zone (which is how we expect the date times from the API gateway to be formatted), it is interpreted as a UTC time, but all the fetch and format functions render the date in the host system's time zone.

From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse:
> For example, "2011-10-10" (date-only form), "2011-10-10T14:48:00" (date-time form),
> or "2011-10-10T14:48:00.000+09:00" (date-time form with milliseconds and time zone)
> can be passed and will be parsed. When the time zone offset is absent, date-only forms
> are interpreted as a UTC time and date-time forms are interpreted as local time.

I've refactored the code a bit to ensure that we're aligning with the underlying UTC storage format. `parseApiGatewayDate()` now looks like this:

```js
export function parseApiGatewayDate(dateString: ApiGatewayDateString): Date {
  return toDate(dateString, { timeZone: pacificTimeZone })
}
```

It takes advantage of `date-fns-tz`'s [`toDate()`](https://github.com/marnusw/date-fns-tz#todate) functionality which allows us to specify the originating time zone for a date time string that does not specify the time zone.

Example from `date-fns-tz`'s docs:
```js
// When there is no offset in the date string the timeZone property is used
const bangkokDate = toDate('2014-10-25T13:46:20', { timeZone: 'Asia/Bangkok' })
```

When there is an offset in the date time string, the `timeZone` option is ignored:
```js
// Since toDate simply clones a Date instance timeZone option is effectively ignored in this case
const date = new Date('2014-10-25T13:46:20Z')
const clonedDate = toDate(date, { timeZone: 'Europe/Paris' })
```

lomky (approved):

Issue #241, opened 2021-06-15
Update to Next.js 11


kalvinwang:
### Description

If it takes less than 1-2 hours total (including getting it to work with dependencies), let's update to Next.js 11 before UAT testing for user-facing launch.

### Acceptance Criteria

- [ ] Next.js 11 running on staging, or close this ticket if it can't be done in 1-2 hours

PR #344, opened 2021-07-29
Support changing the language to Spanish in Storybook and Jest

### LINKED ISSUE #331: Enable Spanish language in Storybook and Jest
### Description

It would be helpful to be able to view Storybook in Spanish to verify the content translation is working as expected. Similarly, it's important that we also test the Spanish content translation, so we need to enable additional locales in jest as well.

### Acceptance Criteria

- [ ] Clicking the language switcher in Storybook changes the language to Spanish
- [ ] Able to test Spanish in Jest

rocketnova:
## Ticket

Resolves #331 

## Changes

- Support changing the language to Spanish in Storybook
  - Merge `.storybook/i18n.js` into `.storybook/preview.js`  
  - Install [storybook-addon-i18next](https://github.com/fynncfchen/storybook-addon-i18next) to support changing the language in Storybook using manager controls
- Support changing the language to Spanish in Jest
  - Split `setupTests.tsx` into two files: `jest-setup.ts` and `jest-i18n.ts`
- Re-enable `useSuspense`

### Storybook

To support languages in Storybook, instead of trying to make the NextJS router change work, we are using the `storybook-addon-i18next` addon, which adds a control for selecting the language.

Screen Shot 2021-07-28 at 5 30 50 PM

As part of this PR, I was able to merge `.storybook/i18n.js` into `.storybook/preview.js`, which resolved an issue where we were calling `i18n.init()` twice. (This warning from the [docs](https://www.i18next.com/overview/api#api) was making me nervous.)

Screen Shot 2021-07-28 at 5 39 29 PM

Because `storybook-addon-i18next` has `react-i18next` as a dependency, it's really important that we make sure we only have one version of `react-i18next` installed, otherwise the locales get split up really weird. See https://github.com/isaachinman/next-i18next/issues/324#issuecomment-888664961

After #241, we may want to revisit this solution, because `storybook-addon-next-router` has had a [recent update](https://github.com/lifeiscontent/storybook-addon-next-router/releases/tag/v3.0.5) that adds support for "support for storybook i18n, so locales work with storybook and next.js"

### Jest

To support languages in Jest, I added the Spanish translation files as resources to the `i18n.init()`. To change the language in a test, we do something like this:

```ts
import i18n from '../jest-i18n'

describe('A test', () => {
  it('that is in Spanish', async () => {
    // Change the language to Spanish
    await i18n.changeLanguage('es')
    ... <test goes here> 
    // Change the language back to English
    await i18n.changeLanguage('en')
  })
})
```

To make that easier, I  split up and renamed `setupTests.tsx` so that `jest-i18n.ts` could export the configured `i18n` instance.

See https://react.i18next.com/misc/testing#testing-without-stubbing for more information about testing i18n without stubbing.

### Suspense

Mostly unrelated to the above changes, I re-enabled `useSuspense` because I came across this in the [react-i18next docs](https://www.i18next.com/overview/api#api):

> In case of react-i18next make sure useSuspense is enabled or handle the ready state in HOCs or hooks yourself.

Re-enabling it doesn't appear to have caused any interference with next-i18next, so I added it to this PR.

### Storybook

While #299 is not resolved, comment out the following lines, then run `yarn storybook`, and click on the globe icon in the top menu bar to select your language https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98

Note that some Spanish translation strings are still missing from `main` at the moment, but will be added with #329.

### Jest

Follow the example above to add a sample test in Spanish. `tests/pages/index.test.tsx` is a good one.


.storybook/main.js
```diff
@@ -1,15 +1,13 @@
...
-  "addons": [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials",
+  stories: ['../stories/**/*.stories.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
+  addons: [
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
     '@storybook/addon-a11y',
-  // Handle SCSS modules
+    'storybook-addon-i18next/register',
```
rocketnova, line 9: Note: This is actually the change to this file. Everything else is just auto-reformatting.

package.json
```diff
@@ -109,6 +111,7 @@
     "**/trim": "^1.0.0",
     "**/prismjs": "^1.24.0",
     "**/glob-parent": "^5.1.2",
-    "**/browserslist": " ^4.16.5"
+    "**/browserslist": " ^4.16.5",
+    "storybook-addon-i18next/react-i18next": "^11.8.8"
```
rocketnova, line 115: Important! You can only have one react-i18next version installed, otherwise you end up with locales stored in different stores unable to access each other. I resolved this by using yarn's [selective resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) feature. See isaachinman/next-i18next#324 (comment)

.storybook/main.js
lomky: thank you!

lomky: I'm a bit suspect that this syntax doesn't match the rest, and I can't see the addon in my local storybook

lomky (commented):
I think we're missing something in the code vs what's installed on your local dev, as it doesn't load the globe for me!

lomky (approved):
worked after a hard reinstall!

.storybook/preview.js
```diff
@@ -45,15 +39,23 @@ export const parameters = {
 export const decorators = [
   (Story, Context) => {
     i18n.use(initReactI18next).init({
-      lng: 'en',
       fallbackLng: 'en',
       ns: ['common', 'claim-details', 'claim-status'],
       defaultNS: 'common',
+      debug: true,
```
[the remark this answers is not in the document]
rocketnova: I copied it over from `.storybook/i18n.js`. It looks like it is for logging info-level console output: https://www.i18next.com/overview/configuration-options#logging

.storybook/main.js
rocketnova: I suspect your comment might have been before you hard-re-installed, but in case it was because the syntax is weird. I did a little more work:

* Turns out that storybook addons are broken up into [3 categories](https://storybook.js.org/docs/react/configure/storybook-addons):
  * essential: ships with storybook
  * core: has the `@storybook` prefix and are located here https://github.com/storybookjs/storybook/tree/next/addons
  * community: everything else; doesn't necessarily follow a naming convention
* Also turns out that I don't need the "/register" portion of the string. This syntax was used before storybook switched over to [declarative config](https://storybook.js.org/blog/declarative-storybook-configuration/). I've removed it.

Issue #347, opened 2021-07-30
Add fallback content for Scenario 2 if no interview time can be shown


rocketnova:
### Description

To handle the error case where the `pendingDetermination.timeSlotDesc` is malformed and unable to be displayed to the user, we need to display alternate content in the `<ClaimSummary>` component.

### Acceptance Criteria

- [ ] `We identified a potential issue that could make you ineligible for benefits. You have a [phone interview scheduled](link to ui home).` for `claim-status:scenarios.scenario2.summary.0.text`

rocketnova:
I discussed this with @kalvinwang yesterday and confirmed that this is actually a non-existent edge case: it's not possible to reach Scenario 2 and then have a malformed `pendingDetermination.scheduleDate`. Instead, if there are issues with malformed data, we should log it. I'll update #320 with a note about that.

Issue #346, opened 2021-07-30
Add final missing translation strings


rocketnova:
### Description

While working on #329, I identified that we're missing a handful of Spanish translation strings. I've notified @nicoleslaw about it and I'm opening this ticket to capture the work.

### Acceptance Criteria

- [x] All remaining Spanish translation strings have been added:
  - [x] "Your Next Steps" — `Sus próximos pasos`
  - [x] "EDD Next Steps" — `Próximos pasos de EDD`
  - [x] Confirm all "EU#" extension types — see [doc](https://docs.google.com/document/d/18ZIc1y8kq9AGxdNCVIyGKWMk5aanBz0p5SaPj2prGPc/edit#)
  - [x] "California Extended Duration (CAL-ED)" — `Duración de la Extensión de California (CAL-ED)`
  - [x] "Interstate Unemployment Insurance" — `Interestatal del Seguro de Desempleo`
  - [x] "NAFTA Transitional Assistance" —`Asistencia de transición del TLCAN`

nicoleslaw:
Hi Rocket, I've updated this doc to match your questions based on EDD's existing content. I will also confirm with their translation team: https://docs.google.com/document/d/18ZIc1y8kq9AGxdNCVIyGKWMk5aanBz0p5SaPj2prGPc/edit#

rocketnova:
> Hi Rocket, I've updated this doc to match your questions based on EDD's existing content. I will also confirm with their translation team: https://docs.google.com/document/d/18ZIc1y8kq9AGxdNCVIyGKWMk5aanBz0p5SaPj2prGPc/edit#

@nicoleslaw You are fantastic. ❤️ Thank you!

PR #345, opened 2021-07-30
Add appointment to Claim Status summary

rocketnova:
## Ticket

Resolves #341 

## Changes

- Create `<Appointment>` component (with story and tests)
- Create `<ClaimSummary>` component (with story and tests)
- Add business logic for when to display an appointment as part of Scenario 2
- Add translation strings for appointments
- Add missing Scenario 2 and Scenario 3 snapshot tests for `<ClaimStatus>` component
- Add test helpers (broken out of `getScenarioContent.test.tsx`)
- Fix bugs in formatDate
- Fix bugs in timeSlot
- Add a util file for handling Sentence case in strings

## Context

This PR is big, but it is only actually handling one very narrow case: when to display the appointment date (`pendingDetermination.scheduleDate`) and time (`pendingDetermination.timeSlotDesc`) and how to format it.

## Testing

`yarn test`

ALSO

1. While #299 is not resolved, comment out the following lines: https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98
2. Run `yarn storybook`
3. Navigate to Page > Default 
4. Select scenario "Determination interview: scheduled"
5. **Verify it looks right**
6. Click the globe in the top menu bar and select "Español"
7. **Verify it looks right**
8. Navigate to Atoms > Appointment
9. Select a date (ignore the time picker) and **verify** in both English and Spanish
10. Enter a start and end time (must be between 1–12) and **verify** in both English and Spanish

Note: that some Spanish translation strings are still missing from `main` at the moment, but will be added with #329.

PR #348, opened 2021-07-30
Integrate Final Content into the Claim Tracker Application

### LINKED ISSUE #329: Integrate Final Content into the Claim Tracker Application

rocketnova:
## Ticket

Resolves #329 

## Changes

- Update final English content
- Update Spanish content
- Add new program & extension types
- Support unexpected program types

## Context

- [x] Final English strings are in the application
  - [x] Common strings
  - [x] scenario specific strings
- [x] Translated strings are in the application

I split out a couple issues from this one:
1. #346
2. #347

## Testing

Follow the testing instructions in #345 and check each scenario (English & Spanish) and each program type (English & Spanish).

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -4,9 +4,8 @@
     "alt-image-edd": "EDD Employment Development Department State of California Home",
     "edd-home": "EDD Home",
     "help": "Help",
-    "welcome": "Welcome, name",
     "logout": "Log Out",
-    "uio-home": "UI Online Home"
+    "uio-home": "UI Home"
   },
   "title": "Claim Tracker",
   "welcome": "Claim Tracker",
@@ -23,8 +22,7 @@
     "claim-balance": "Claim Balance",
     "weekly-benefit-amount": "Weekly Benefit Amount",
     "last-payment-issued": "Last Payment Issued",
-    "extension-type": "Extension Type",
-    "extension-end-date": "Extension End Date"
+    "extension-type": "Extension Type"
   },
   "footer": {
     "toTop": "Back to Top",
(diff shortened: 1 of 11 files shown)


public/locales/en/common.json
```diff
@@ -36,7 +34,7 @@
   },
   "timeout-modal": {
     "header": "Your Session Will End Soon",
-    "warning": "To protect your information, you will be logged out in {{count}} minute if you do not continue",
+    "warning": "To protect your information, you will be logged out in {{count}} minute if you do not continue.",
```
nicoleslaw, line 37: Should this be minutes, or are there times when it would be 1 minute?

kalvinwang: Let's just go with plural minutes here since it's only singular for... one minute

lomky: we _should_ be able to use [i18n pluralization](https://www.i18next.com/translation-function/plurals) but it was giving me trouble. It might be worth another try though since Rocket solved a bug in my i18n setup!

lomky: wait, there is the plural down below on the next line! it properly pluralizes:

image

image

lomky (approved):
I defer to @nicoleslaw for final okay on the content itself, but all the tech seems to work right with the new strings

public/locales/en/common.json
kalvinwang: Oh cool, didn't realize we pluralized, thanks for checking Kat!

kalvinwang (approved):
lgtm!

nicoleslaw (approved):
I don't have any other changes for this but am assuming some of this will be updated via https://github.com/cagov/ui-claim-tracker/issues/346

PR #358, opened 2021-08-05
Fix pino-applicationinsights in storybook

### LINKED ISSUE #299: Installing pino-applicationinsights has broken storybook builds

rocketnova:
## Ticket

Resolves #299 

## Changes

- Mocks the `pino-applicationinsights` module in storybook
- Installs the [`node-noop`](https://github.com/euank/node-noop) package

## Context

When I added `pino-applicationinsights` to the repo to handle logging directly to Azure Application Insights, it broke Storybook builds (locally and on Chromatic). We were able to isolate that if we commented out the following lines, storybook builds would work successfully: https://github.com/cagov/ui-claim-tracker/blob/32990f892c6b84017c30af5d273fdd94410649e7/pages/index.tsx#L95-L98

The specific storybook build errors looked like this (for many many modules):

```sh
Module not found: Error: Can't resolve 'async_hooks' in '<snip>/ui-claim-tracker/node_modules/cls-hooked'
```

This is [apparently what the error looks like](https://github.com/elastic/apm-agent-nodejs/issues/1075#issuecomment-494587021) when "when trying to bundle Node.js code that uses non-browser-compatible code without explicitly telling webpack to allow it". An [extremely helpful](https://github.com/wallabyjs/public/issues/2525#issuecomment-700326595) comment in the wallabyjs github repo suggested using webpack's [NormalModuleReplacementPlugin](https://webpack.js.org/plugins/normal-module-replacement-plugin/) to replace the module with a mock. And the [wallaby docs](https://wallabyjs.com/docs/integration/webpack.html#normalmodulereplacement-plugin) suggested using the `node-noop` module:

```js
module.exports = function (wallaby) {
  var wallabyPostprocessor = wallabyWebpack({
    ...
      plugins: [
        new webpack.NormalModuleReplacementPlugin(/\.(gif|png|scss|css)$/, 'node-noop')
      ]
    ...
  });
};
```

## Testing

1. Checkout this branch
2. `yarn storybook`
3. Storybook should run with no errors

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const webpack = require('webpack')
 
 module.exports = {
   stories: ['../stories/**/*.stories.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
@@ -25,11 +26,15 @@ module.exports = {
     },
   ],
   webpackFinal: (config) => {
+    // applicationinsights is meant for node.js applications, storybook uses a browser environment.
+    // Use webpack's NormalModuleReplacementPlugin to mock the applicationinsights in storybook.
+    // See https://webpack.js.org/plugins/normal-module-replacement-plugin
+    config.plugins.push(new webpack.NormalModuleReplacementPlugin(/pino-applicationinsights/, 'node-noop'))
     return {
       ...config,
       node: {
         ...config.node,
-        fs: 'empty', //required to with with next-i18next
+        fs: 'empty', // required for next-i18next
       },
     }
   },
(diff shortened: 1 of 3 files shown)


kalvinwang (approved):
Thanks for the super detailed PR description, nice spelunking!

PR #363, opened 2021-08-06
Only connect to the API gateway if there is a unique number

### LINKED ISSUE #334: Only ping API gateway when unique number header exists
### Description

We should only send requests to the API gateway when a unique number header exists, in all environments. 

### Acceptance Criteria

- [ ] Confirm that requests aren't being sent to the API Gateway unless a unique number exists in the header

rocketnova:
## Ticket

Resolves #334 

## Changes

- Update `index.tsx` to only query the API gateway if there is a unique number in the header

## Context

We don't want to be querying the API gateway for healthchecks and error edge cases. 

## Testing

Compare Application Insights between an environment where this branch has been deployed (Internal) and an environment where `main` has been deployed (Staging) and confirm when the `Application error` error message is logged and not logged vs when the `Missing unique number` error message is logged/not logged.

_Note:_ Testing this in remote environment is not currently possible due to some Application Insights issues.


tests/utils/queryApiGateway.test.tsx
```diff
@@ -195,13 +195,31 @@ describe('Building the API url', () => {
...
     expect(id).toStrictEqual(goodRequest.headers.id)
+
+    // Restore env vars
+    restore()
   })
 
-  it('is not case sensitive', () => {
-    const key = 'uniqueNumber'
+  // Each test case should be:
+  // [test description, mock env var
```
lomky, line 211:
```suggestion
  // [test description, mock env var]
```

tests/utils/queryApiGateway.test.tsx
```diff
@@ -195,13 +195,31 @@ describe('Building the API url', () => {
...
+  ]
+  it.each(caseSensitiveTestCases)('is not case sensitive: %s', (description: string, key: string) => {
+    // Mock process.env
+    const restore = mockEnv({
+      ID_HEADER_NAME: key,
+    })
+
     const request = {
       headers: {
         uniquenumber: 'lower',
```
lomky, line 225: Should this be using `key` or `id`? I don't quite follow the setup steps of setting the env, then building the request, then running the test.

lomky (commented):
Couple questions around the test flow!

tests/utils/queryApiGateway.test.tsx
rocketnova: This test builds a different `request` object, instead of using `goodRequest` like in the other tests in this file. This test ensures that the `ID_HEADER_NAME` environment variable is **not** case sensitive. The setup for the test sets up the `ID_HEADER_NAME` env var with three different **case-sensitive** versions of the string `uniquenumber` and ensures that they all have the same behavior.

kalvinwang: nice!

PR #366, opened 2021-08-10
Revert changes from PR #288

kalvinwang:
#288 caused requests to stop being logged in Application Insights (though not traces). This is a temporary fix while we investigate the root cause.


rocketnova (approved):
🚢

PR #374, opened 2021-08-11
Fix unserializable json error in Scenario 2

### LINKED ISSUE #372: Correct Scenario 2 json issue
### Description

For an unknown reason (possibly divergent versions of node), my local machine is returning an unserializable JSON error for Scenario 2 when passing a Date object rather than a string for appointment date. We should pass a date string instead of a Date object.

### Acceptance Criteria

- [ ] Pass date string for appointment date instead of date object

rocketnova:
## Ticket

Resolves #372 

## Changes

- Change the `AppointmentContent.date` type from a `Date` object to a `string` representation of the date

## Context

For Scenario 2, on my local development, I was receiving the following error:

```
SerializableError: Error serializing `.scenarioContent.statusContent.summary.appointment.date` returned from `getServerSideProps` in "/".
Reason: `object` ("[object Date]") cannot be serialized as JSON. Please only return JSON serializable data types.
```

This error was not present on the remote environment for the same branch (`main`). I suspect this may be because I'm running a newer version of nodejs than Azure. If that's true, this PR helps future-proof the app. If not, it resolves the issue so hopefully it will not occur in prod and non-prod environments.

**Note: I merged #379 into this PR, so this PR contains both the unserializable json error and the bold font fix:**
- Use Google fonts for Source Sans Pro: bold, italic, bolditalic
- Remove local bold ttf file

## Testing

Use modHeader to test locally against Scenario 2.


kalvinwang (approved):
Lgtm, thanks! Were the last six commits intentionally on this branch? I assume they’re meant for another PR? Not a big deal if you want to just update this PR description either

lomky (approved):
I can't actually load Scenario 2 locally anymore (the date has past, so now it's scenario 3!). The code looks fine, the tests pass. I would suggest cleaning out the unrelated font changes from this branch though!

lomky (approved):
We have a new UniqueNum for scenario 2! And I reproduced the error on main and this branch resolves it!

PR #371, opened 2021-08-11
Support null ClaimDetails response

lomky:
Changes:
 - Updates `getScenarioContent` to handle null claimDetails
 - Updates `ClaimSection` component to display/not display Claim Details based on that
 - Updates the ApiStub to support null claimDetails
 - Adds tests
 - Updates storybook

To Test:
  - `yarn test`
  - Open storybook, open the `Page` component & toggle `hasClaimDetails` to false

Resolves #364 

AC:

 - [x] When the API returns a null claimDetails object, hide the Claim Details section (including the heading and the fields it contains) from the Claim Tracker page

PR #375, opened 2021-08-12
Hide fields in the Claim Details section if the field returns null

### LINKED ISSUE #365: Hide fields in the Claim Details section if the field returns null
### Description

There are certain use cases in which it's possible for certain Claim Details fields to return null:
- `Claim Balance`
- `Weekly Benefit Amount`
- `Last Payment Issued` 
- `Last Payment Amount`

In order to fail gracefully in those cases, we should implement logic that hides each individual Claim Details field if the field returns null for a given claimant.

Note that in order to display Last Payment Issued to the user, we combine data from two API fields  `Last Payment Issued` and  `Last Payment Amount` and display them in a single field. In this case, if one of the source fields is null and the other isn't, we should still hide the whole field as displayed in Claim Details.  

### Acceptance Criteria

- [ ] If a field in the `claimDetails` object returns null, the corresponding display field in the Claim Details section is hidden

lomky:
- Makes the three fields indicated in #365 safely nullable
  - Hides the entire field when their value is null
  - Basic tests for the API parsing and the Component display
  - Storybook tweaks
  - Removes an extraneous type change (see https://github.com/cagov/ui-claim-tracker/pull/371#discussion_r687879970)

Testing:
 - Load storybook and look at the ClaimDetails component - delete the string and observe that the field disappears
 - `yarn test`

===

Resolves #365

- [x] If `Claim Balance`, `Weekly Benefit Amount`, `Payment Issued`, `Last Payment Amount` in the claimDetails object returns null, the corresponding display field in the Claim Details section is hidden

### DIFF
--- a/stories/ClaimDetails.stories.tsx
+++ b/stories/ClaimDetails.stories.tsx
@@ -49,6 +49,6 @@ ClaimDetails.args = {
   benefitYear: '3/21/2020 - 3/20/2021',
   claimBalance: '$508.00',
   weeklyBenefitAmount: '$120.00',
-  lastPaymentIssued: '4/29/2021',
+  lastPaymentIssued: '$120.00 on 4/29/21',
   extensionType: 'Tier 2 Extension',
 }
--- a/stories/ClaimDetails.stories.tsx
+++ b/stories/ClaimDetails.stories.tsx
@@ -7,6 +7,38 @@ export default {
   title: 'Component/Page Section/Claim Details',
   component: ClaimDetailsComponent,
   decorators: [withNextRouter],
+  argTypes: {
+    programType: {
+      control: {
+        type: 'text',
+      },
+    },
+    benefitYear: {
+      control: {
+        type: 'text',
+      },
+    },
+    claimBalance: {
+      control: {
+        type: 'text',
+      },
+    },
+    weeklyBenefitAmount: {
+      control: {
+        type: 'text',
+      },
+    },
+    lastPaymentIssued: {
+      control: {
+        type: 'text',
+      },
+    },
+    extensionType: {
+      control: {
+        type: 'text',
+      },
+    },
+  },
 } as Meta
 
 const Template: Story<ClaimDetailsProps> = (args) => <ClaimDetailsComponent {...args} />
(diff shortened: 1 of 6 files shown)


stories/ClaimDetails.stories.tsx
```diff
@@ -7,6 +7,38 @@ export default {
...
+    },
+    extensionType: {
+      control: {
+        type: 'text',
+      },
+    },
+  },
 } as Meta
 
 const Template: Story<ClaimDetailsProps> = (args) => <ClaimDetailsComponent {...args} />
```
rocketnova, line 44: This is definitely not a result of your PR, but while reviewing I noticed that the `lastPaymentIssued` story arg should probably be `'$120.00 on 4/29/21'` or something like that.

tests/components/ClaimDetails.test.tsx
```diff
@@ -0,0 +1,105 @@
+import renderer from 'react-test-renderer'
+import { render, screen } from '@testing-library/react'
+import getScenarioContent, { ScenarioType } from '../../utils/getScenarioContent'
+import apiGatewayStub from '../../utils/apiGatewayStub'
+import { ScenarioContent, ClaimDetailsContent } from '../../types/common'
+
+import { useRouter } from 'next/router'
```
rocketnova, line 7: I'm also seeing these linting issues locally.

tests/utils/getClaimDetails.test.tsx
```diff
@@ -68,4 +68,82 @@ describe('Constructing the Claim Details object', () => {
     const claimDetails = getClaimDetails(rawDetails)
     expect(claimDetails).toStrictEqual(expected)
   })
+
```
rocketnova, line 71: Thanks for adding these component tests!

types/common.tsx
```diff
@@ -44,7 +44,6 @@ export interface Claim {
   claimDetails?: null | ClaimDetailsResult
   hasPendingWeeks?: null | undefined | boolean
   hasCertificationWeeksAvailable?: null | undefined | boolean
-  hasClaimDetails?: null | undefined | boolean
```
rocketnova, line 47: Thank you!

rocketnova (approved):
Approved with some comments!

PR #380, opened 2021-08-12
Scenario 1: Handle invalid schedule date and invalid request date

### LINKED ISSUE #370: Handle invalid schedule date for determining Scenario1
### Description

Scenario 1 expects there to be:

  - `null` `determinationStatus`
  - `null` `scheduleDate`
  - a `requestDate`

Instead, it should consider:

  - `null` determinationStatus
  - invalid scheduleDate 
    - i.e. either `null` and `0001-01-01T00:00:00`
  - a requestDate
    - we should confirm if this needs a validity check 

Line of code implicated:
https://github.com/cagov/ui-claim-tracker/blob/50644ed299f3ed4136b93552680fb76ee71e2093/utils/getScenarioContent.tsx#L125

### Acceptance Criteria

- [ ] When we receive `scheduleDate` the value `0001-01-01T00:00:00`, we should consider that equivalent to `null`
- [ ] We have confirmed `requestDate` needs a validity check or not, and handled it if it does

rocketnova:
## Ticket

Resolves #370 

## Changes

- Add a function `isDateStringFalsy()` to check if a dateString evaluates to JS falsy or equals `0001-01-01T00:00:00`
- Require `pendingDetermination.requestDate` to be a valid date
- Update Scenario 1 to use `isDateStringFalsy()` to validate `pendingDetermination.scheduleDate`

## Context

This PR adds two additional validations for the way we handle dates in Scenario 1:
1. `requestDate` must be a valid date
2. `scheduleDate` must either evaluate to false (null, empty, undefined) or equal `0001-01-01T00:00:00`

## Testing

`yarn test`

### DIFF
--- a/tests/utils/getScenarioContent.test.tsx
+++ b/tests/utils/getScenarioContent.test.tsx
@@ -3,6 +3,7 @@ import MockDate from 'mockdate'
 import { getMockPendingDetermination, getPendingDeterminationWithScheduleDate } from '../testHelpers'
 import { PendingDetermination } from '../../types/common'
 import apiGatewayStub from '../../utils/apiGatewayStub'
+import { formatFromApiGateway } from '../../utils/formatDate'
 import {
   getScenario,
   identifyPendingDeterminationScenario,
@@ -74,7 +75,7 @@ describe('The determination interview is not yet scheduled (scenario 1) when all
     const pendingDetermination = getMockPendingDetermination()
     pendingDetermination.determinationStatus = null
     pendingDetermination.scheduleDate = null
-    pendingDetermination.requestDate = 'any non-null value'
+    pendingDetermination.requestDate = formatFromApiGateway(-30)
     const result = identifyPendingDeterminationScenario([pendingDetermination])
     expect(result.scenarioType).toBe(ScenarioType.Scenario1)
   })
@@ -83,7 +84,7 @@ describe('The determination interview is not yet scheduled (scenario 1) when all
     const pendingDetermination = getMockPendingDetermination()
     pendingDetermination.determinationStatus = ''
     pendingDetermination.scheduleDate = null
-    pendingDetermination.requestDate = 'any non-null value'
+    pendingDetermination.requestDate = formatFromApiGateway(-30)
     const result = identifyPendingDeterminationScenario([pendingDetermination])
     expect(result.scenarioType).toBe(ScenarioType.Scenario1)
   })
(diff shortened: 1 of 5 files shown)


tests/utils/getScenarioContent.test.tsx
```diff
@@ -74,7 +75,7 @@ describe('The determination interview is not yet scheduled (scenario 1) when all
     const pendingDetermination = getMockPendingDetermination()
     pendingDetermination.determinationStatus = null
     pendingDetermination.scheduleDate = null
-    pendingDetermination.requestDate = 'any non-null value'
+    pendingDetermination.requestDate = formatFromApiGateway(-30)
```
rocketnova, line 78: Note: requestDate must now be a valid date, so these tests pick a random date in the past.

utils/getScenarioContent.tsx
```diff
@@ -122,8 +122,8 @@ export function identifyPendingDeterminationScenario(
     // AND requestDate has a value
     else if (
       !pendingDetermination.determinationStatus &&
-      !pendingDetermination.scheduleDate &&
-      pendingDetermination.requestDate
+      isValidDate(pendingDetermination.requestDate) &&
+      (!pendingDetermination.scheduleDate || isNullDateString(pendingDetermination.scheduleDate))
```
rocketnova, line 126: Note: This is the core logic being updated in this PR.

utils/formatDate.ts
```diff
@@ -91,6 +91,13 @@ export function isValidDate(dateString: ApiGatewayDateString): boolean {
   }
 }
 
+/**
+ * Returns true if the dateString is Jan 1, year 1.
+ */
+export function isNullDateString(dateString: ApiGatewayDateString): boolean {
+  return dateString === '0001-01-01T00:00:00'
+}
```
lomky, line 99: maybe a nitpick, but I thought this would check `'0001-01-01T00:00:00' | null` based on the name - should it be that? That would also let you simplify the call in `getScenarioContent`

lomky (approved):
One optional suggestion!

utils/formatDate.ts
kalvinwang: Not necessary, but what if this was isPre2020DateString instead? it seems like you've added all the other code for that already in this PR. and the TODO below could be just for the logging part.

kalvinwang (approved):
lgtm!

utils/formatDate.ts
rocketnova: I think this should actually be turned into something like `isDateStringFalsy()` because the logic for Scenario 1 is that `scheduleDate` must be null. To be extra safe, I'm checking for falsy (empty string, null, undefined) and want to extend `0001-01-01T00:00:00` to be included in the definition of "falsy" in this case.

utils/formatDate.ts
```diff
@@ -92,10 +92,12 @@ export function isValidDate(dateString: ApiGatewayDateString): boolean {
 }
 
 /**
- * Returns true if the dateString is Jan 1, year 1.
+ * Returns true if the dateString is falsy.
+ *
+ * We are defining falsy here to mean: null, empty string, undefined, or 0001-01-01T00:00:00.
```
lomky, line 97:
```suggestion
 * We are defining falsy here to mean: standard JS falsy (null, empty string, undefined, etc) or 0001-01-01T00:00:00.
```

lomky (approved):
🎉

Issue #338, opened 2021-07-26
Validate json strings on build


lomky:
Right now we read strings our of our i18n locale files and our urls.json file. The code does not know whether these strings exist, leaving us with a bunch of `|| null` types.

We know all of the lookups at build time, so we should be able to validate that!

_Reference: @rocketnova in https://github.com/cagov/ui-claim-tracker/pull/333#r676842841_

PR #397, opened 2021-08-19
Look for environment-specific UIO landing URLs first

kalvinwang:
This PR looks for a new environment variable to use as the UIO landing URL. If not found, it falls back to the current production/public UIO landing URL.

After this PR is merged, I'll add the SIT07 landing URLs to our test environment and test there

===

- [x] Relevant documentation (e.g. READMEs, Technical Foundation) updated


components/TransLine.tsx
```diff
@@ -15,7 +15,11 @@ export interface TransLineProps extends TransLineContent {
 function resolveUrl(link: I18nString, userArrivedFromUioMobile: boolean) {
   // Special case for UIO homepage links.
   if (link === 'uio-home') {
-    const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
+    // Optional environment-specific links back to the UIO landing page, used by EDD testing
+    const uioLandingUrl = process.env.URL_UIO_LANDING ?? getUrl('uio-home-url-desktop')
+    const uioMobileLandingUrl = process.env.URL_UIOMOBILE_LANDING ?? getUrl('uio-home-url-mobile')
+
+    const uioHomeLink = userArrivedFromUioMobile ? uioMobileLandingUrl : uioLandingUrl
     if (uioHomeLink) {
```
kalvinwang, line 23: @rocketnova I don't understand this if statement-- is it to protect against the possibility of uio-home-url-desktop/mobile, not existing as a string? If they do exist, then uioHomeLink will always be true right?

README.md
```diff
@@ -26,6 +26,7 @@ relates to weekly certification
 - CERTIFICATE_DIR: The path to the client certificate (certificate must be in PFX/P12 format)
 - PFX_FILE: The name of the client certificate file
 - (Optional) PFX_PASSPHRASE: The import passphrase for the client certificate if there is one
+- (Optional) URL_UIO_LANDING, URL_UIOMOBILE_LANDING: The environment-specific links to the UIO landing page
```
lomky, line 29:
```suggestion
- (Optional) URL_UIO_LANDING, URL_UIOMOBILE_LANDING: The environment-specific links to the UIO landing pages
```

components/TransLine.tsx
```diff
@@ -15,7 +15,11 @@ export interface TransLineProps extends TransLineContent {
 function resolveUrl(link: I18nString, userArrivedFromUioMobile: boolean) {
   // Special case for UIO homepage links.
   if (link === 'uio-home') {
-    const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
+    // Optional environment-specific links back to the UIO landing page, used by EDD testing
+    const uioLandingUrl = process.env.URL_UIO_LANDING ?? getUrl('uio-home-url-desktop')
+    const uioMobileLandingUrl = process.env.URL_UIOMOBILE_LANDING ?? getUrl('uio-home-url-mobile')
+
+    const uioHomeLink = userArrivedFromUioMobile ? uioMobileLandingUrl : uioLandingUrl
```
rocketnova, line 22: Nit: Not necessary for launch, but I think this is enough logic to put it somewhere else (like in `getUrl.ts`) rather than in the component.

components/TransLine.tsx
lomky: Typescript will yell if we don't narrow here, because the code doesn't know those json strings _definitely_ exist - we have a ticket to see if there's some way to assert the json strings do exist #338

lomky (approved):

components/TransLine.tsx
rocketnova: `getUrl()` can return undefined. I'm not positive what happens if `public/urls.json` disappears or is malformed or has a read-error, but there's definitely the possibility that `public/urls.json` exists and the `linkKey` passed in doesn't exist in the file.

rocketnova (approved):
LGTM!

components/TransLine.tsx
rocketnova: > Typescript will yell if we don't narrow here, because the code doesn't know those json strings _definitely_ exist - we have a ticket to see if there's some way to assert the json strings do exist #338

Agreed that we should know at build time, but I think the check is important in case something horrible and unexpected happens at runtime

kalvinwang: Gotcha, this all makes sense. Thanks!

components/TransLine.tsx
kalvinwang: Oh yeah, I should have put it there. Moved!

PR #398, opened 2021-08-20
Adds luxon DateTimes & uses them inside critical functions

### LINKED ISSUE #385: Scenario 2
### Acceptance Criteria

- [ ]

lomky:
PR 1 of 2

This is the minimal changes to actually solve #385, while leaving the rest of the date functionality unchanged.

 - Adds luxon DateTime
 - Sets our Luxon TZ to 'America/Los_Angeles'
 - Teaches `formatFromApiGateway` to act properly around midnight
 - Teaches `isDatePast` to act properly around midnight

However this is a _bad idea_ to do halfway! now we have two time-handling processes! That handle time zones differently! AHHH

So see the next PR in the stack for the full refactor. I just want to hold this distinct for reviewing.

Testing:
  - [ ] All scenarios still load properly - but especially 2 & 3
  - [ ] Tests pass, with no snapshot changes

===

Resolves #385

- [x] Scenario 2 should be returned for appointments that are "today" until 11:59pm PT. Scenario 3 should not be displayed until it is 12:00am PT the next calendar day

### DIFF
--- a/tests/utils/formatDate.test.ts
+++ b/tests/utils/formatDate.test.ts
@@ -1,7 +1,14 @@
 import MockDate from 'mockdate'
 import { toDate } from 'date-fns-tz'
+import { DateTime, Settings } from 'luxon'
 
-import formatDate, { formatAppointmentDate, isDatePast, isDateStringFalsy, isValidDate } from '../../utils/formatDate'
+import formatDate, {
+  formatAppointmentDate,
+  formatFromApiGateway,
+  isDatePast,
+  isDateStringFalsy,
+  isValidDate,
+} from '../../utils/formatDate'
 
 // Test isValidDate()
 describe('Valid dates: A date is', () => {
@@ -86,5 +93,30 @@ describe('Formatting dates', () => {
     const notFormatted = '2013-09-27T00:00:00'
     const formatted = formatDate(notFormatted)
     expect(formatted).toEqual('9/27/2013')
+  })
+})
+
+// Test formatFromApiGateway()
+describe('Requesting a date adjusted from existing in the correct format', () => {
+  it('has our time set to PT!', () => {
+    expect(DateTime.local().zoneName).toEqual('America/Los_Angeles')
+  })
+
+  it('displays the expected date string', () => {
+    const baseDate = DateTime.fromISO('2021-09-10T02:00:00')
+    const adjustment = -1
+    const newDate = formatFromApiGateway(adjustment, baseDate)
+
+    expect(newDate).toEqual('2021-09-09T00:00:00')
+  })
+
+  it('handles PT edge cases', () => {
+    // Tricky time - 10pm PT means its always* the next day ET & UTC
+    // *like, casually true but don't hold me to this time is hard okay?
+    const baseDate = DateTime.fromISO('2021-09-01T22:00:00')
+    const adjustment = 3
+    const newDate = formatFromApiGateway(adjustment, baseDate)
+
+    expect(newDate).toEqual('2021-09-04T00:00:00')
   })
 })
--- a/utils/formatDate.ts
+++ b/utils/formatDate.ts
@@ -16,11 +16,14 @@ import { isValid } from 'date-fns'
 import { format, toDate, utcToZonedTime } from 'date-fns-tz'
 import enUS from 'date-fns/locale/en-US'
 import es from 'date-fns/locale/es'
+import { DateTime, Settings } from 'luxon'
 
 import { ApiGatewayDateString } from '../types/common'
 
 const pacificTimeZone = 'America/Los_Angeles'
 const apiGatewayFormat = "yyyy-MM-dd'T'HH:mm:ss"
+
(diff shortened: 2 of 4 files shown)


lomky:
The tests are failing because I didn't update the tests until to follow on - if we decide to merge this one without the other I'll backport it

utils/formatDate.ts
```diff
@@ -16,12 +16,15 @@ import { isValid } from 'date-fns'
...
 import enUS from 'date-fns/locale/en-US'
 import es from 'date-fns/locale/es'
+import { DateTime, Settings } from 'luxon'
 
 import { ApiGatewayDateString } from '../types/common'
 
 const pacificTimeZone = 'America/Los_Angeles'
 const apiGatewayFormat = "yyyy-MM-dd'T'HH:mm:ss"
 
+Settings.defaultZone = 'America/Los_Angeles'
```
rocketnova, line 26: Nit:

```suggestion
Settings.defaultZone = pacificTimeZone
```

utils/formatDate.ts
```diff
@@ -48,16 +51,12 @@ export function parseApiGatewayDate(dateString: ApiGatewayDateString): Date {
...
- * currently PST (-8) or PDT (-7).
  */
-export function formatFromApiGateway(daysOffset = 1): string {
-  const today = new Date()
-  today.setDate(today.getDate() + daysOffset)
-  today.setUTCHours(8, 0, 0, 0)
-  return format(today, apiGatewayFormat)
+export function formatFromApiGateway(daysOffset = 1, today = DateTime.now()): string {
+  const newDate = today.plus({ days: daysOffset }).startOf('day')
+  return newDate.toFormat(apiGatewayFormat)
```
rocketnova, line 59: 🎉

tests/utils/formatDate.test.ts
```diff
@@ -88,3 +95,28 @@ describe('Formatting dates', () => {
     expect(formatted).toEqual('9/27/2013')
   })
 })
+
+// Test formatFromApiGateway()
+describe('Requesting a date adjusted from existing in the correct format', () => {
+  it('has our time set to PT!', () => {
+    expect(DateTime.local().zoneName).toEqual('America/Los_Angeles')
```
rocketnova, line 102: Question about luxon: Do you need to also add `Settings.defaultZone = 'America/Los_Angeles'` to this file? How does `DateTime` now which timezone is local?

tests/utils/formatDate.test.ts
```diff
@@ -88,3 +95,28 @@ describe('Formatting dates', () => {
...
+    const baseDate = DateTime.fromISO('2021-09-10T02:00:00')
+    const adjustment = -1
+    const newDate = formatFromApiGateway(adjustment, baseDate)
+
+    expect(newDate).toEqual('2021-09-09T00:00:00')
+  })
+
+  it('handles PT edge cases', () => {
+    // Tricky time - 10pm PT means its always* the next day ET & UTC
+    // *like, casually true but don't hold me to this time is hard okay?
```
rocketnova, line 115:
```suggestion
    // Tricky time - 10pm PT means it's always* the next day ET & UTC
    // *like, casually true but don't hold me to this; time is hard okay?
```

tests/utils/formatDate.test.ts
```diff
@@ -1,7 +1,14 @@
 import MockDate from 'mockdate'
 import { toDate } from 'date-fns-tz'
+import { DateTime, Settings } from 'luxon'
```
rocketnova, line 3: I'm getting a linting warning that `Settings` is imported but not used

rocketnova (approved):
A couple small questions. This looks good and we should 🚢 , but I think we should test Scenario 2 against dev on 8/25 to ensure it's working as expected on staging. I'll add a reminder on my calendar for next week

utils/formatDate.ts
lomky: good nit - but we end up not using the `pacificTimeZone` anywhere else so I just delete it in the follow up!

tests/utils/formatDate.test.ts
lomky: Yup! 😅 cleaned up in next PR!

tests/utils/formatDate.test.ts
lomky: Good question! I tested adding it in various places. Initially I was hopeful adding it to the ServerSide function in `page/index.tsx` would be sufficient, but no. Then I added it to here, also didn't work, and then I added it to utils, and it does work. I believe the Setting affects the behavior everywhere once its set

I think if I were more JS-Server-Stuff savvy I would know a better place to put it 🤔

rocketnova: I feel a little surprised by that behavior 🤔  but 🤷

lomky:
Nooo idea why the locale isn't being respected on the CI here - it works locally 🤔 Feel free to debug while I'm gone, otherwise I'll pick up Thursday!

PR #400, opened 2021-08-20
Support more environment-specific links

kalvinwang:
This PR:

1) makes the UI Home link open in the same window instead of a new tab,
2) makes the new claim link environment specific, and
3) at the end of 30 minutes, redirects to the env specific login link

===

Resolves #XXX

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ relates to weekly certification
 - CERTIFICATE_DIR: The path to the client certificate (certificate must be in PFX/P12 format)
 - PFX_FILE: The name of the client certificate file
 - (Optional) PFX_PASSPHRASE: The import passphrase for the client certificate if there is one
-- (Optional) URL_UIO_LANDING, URL_UIOMOBILE_LANDING: The environment-specific links to the UIO and UIO Mobile landing pages
+- (Optional) URL_PREFIX_UIO, URL_PREFIX_UIO_MOBILE, URL_PREFIX_BPO: Environment-specific path prefixes for UIO and BPO links
 
 For local development:
 
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -14,7 +14,7 @@ export const Header: React.FC<HeaderProps> = ({ userArrivedFromUioMobile = false
   // Return a link back to:
   //   UIO Mobile landing page if user arrived from UIO Mobile
   //   main UIO landing page if user arrived from main UIO
-  const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-home-url-mobile') : getUrl('uio-home-url-desktop')
+  const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-mobile-home-url') : getUrl('uio-home-url')
 
   return (
     <header className="header border-bottom border-secondary">
@@ -34,10 +34,10 @@ export const Header: React.FC<HeaderProps> = ({ userArrivedFromUioMobile = false
                 <span className="text">{t('header.edd-home')}</span>
               </Nav.Link>
             </Navbar.Collapse>
-            <Nav.Link target="_blank" rel="noopener noreferrer" href={getUrl('edd-help-new-claim')}>
+            <Nav.Link target="_blank" rel="noopener noreferrer" href={getUrl('uio-help-new-claim')}>
               <span className="text">{t('header.help')}</span>
             </Nav.Link>
-            <Nav.Link href={getUrl('edd-log-out')}>
+            <Nav.Link href={getUrl('bpo-log-out')}>
               <span className="text">{t('header.logout')}</span>
             </Nav.Link>
           </Nav>
@@ -57,7 +57,7 @@ export const Header: React.FC<HeaderProps> = ({ userArrivedFromUioMobile = false
           <Navbar.Toggle aria-controls="basic-navbar-nav" />
           <Navbar.Collapse>
             <Nav>
-              <Nav.Link target="_blank" rel="noopener noreferrer" href={uioHomeLink}>
+              <Nav.Link rel="noopener noreferrer" href={uioHomeLink}>
                 <span className="text">{t('header.uio-home')}</span>
               </Nav.Link>
             </Nav>
--- a/components/TimeoutModal.tsx
+++ b/components/TimeoutModal.tsx
@@ -55,7 +55,12 @@ export const TimeoutModal: React.FC<TimeoutModalProps> = ({ timedOut, userArrive
     // And at the end, send back to EDD
     setTimeout(() => {
       if (typeof window !== 'undefined') {
-        const eddLoginLink = getUrl('edd-log-in')?.concat(encodeURIComponent(window.location.toString()))
+        // Note that the concatenated portion of this link is functionally useless, as IDM is not currently
+        // able to redirect based on the resource_url parameter concatenated.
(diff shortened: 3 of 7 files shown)


utils/getUrl.ts
```diff
@@ -11,12 +11,22 @@ export type UrlType = keyof typeof urls
...
-  if (linkKey === 'uio-home-url-desktop' && process.env.URL_UIO_LANDING) return process.env.URL_UIO_LANDING
-  if (linkKey === 'uio-home-url-mobile' && process.env.URL_UIOMOBILE_LANDING) return process.env.URL_UIOMOBILE_LANDING
-
   // Explicitly cast to one of the allowed keys in urls.json.
   // If the key does not exist in urls.json, this function will return undefined.
   const key = linkKey as UrlType
+
+  // Optional environment-specific links back to the UIO landing page, used by EDD testing
+  if (key.startsWith('uio') && process.env.URL_PREFIX_UIO) {
+    return urls[key].replace('uio.edd.ca.gov/UIO', process.env.URL_PREFIX_UIO)
```
kalvinwang, line 20: I was thinking of moving these into urls.json but that doesn't seem quite right either...

public/urls.json
```diff
@@ -5,9 +5,9 @@
   "edd-about-contact": "https://www.edd.ca.gov/About_EDD/Contact_EDD.htm",
   "edd-about-privacy-policy": "https://www.edd.ca.gov/About_EDD/Privacy_Policy.htm",
   "edd-ca-gov": "https://edd.ca.gov",
-  "edd-help-new-claim": "https://uio.edd.ca.gov/UIO/Pages/Public/help/index.htm#t=en-US/Public/NewClaim/UIOnlineNewClaimLandingPage.htm",
-  "edd-log-in": "https://portal.edd.ca.gov/WebApp/Login?resource_url=",
-  "edd-log-out": "https://portal.edd.ca.gov/WebApp/Logout",
-  "uio-home-url-desktop": "https://uio.edd.ca.gov/UIO/Pages/Public/ExternalUser/UIOnlineLandingPage.aspx",
-  "uio-home-url-mobile": "https://uiom.edd.ca.gov/UIOM/Pages/Public/ExternalUser/UIOMobileLandingPage.aspx"
+  "bpo-log-in": "https://portal.edd.ca.gov/WebApp/Login",
+  "bpo-log-out": "https://portal.edd.ca.gov/WebApp/Logout",
```
rocketnova, line 9: Nit:
```suggestion
  "bpo-login": "https://portal.edd.ca.gov/WebApp/Login",
  "bpo-logout": "https://portal.edd.ca.gov/WebApp/Logout",
```

utils/getUrl.ts
```diff
@@ -11,12 +11,22 @@ export type UrlType = keyof typeof urls
...
+  }
+
+  if (key.startsWith('uio-mobile') && process.env.URL_PREFIX_UIO_MOBILE) {
+    return urls[key].replace('uiom.edd.ca.gov/UIOM', process.env.URL_PREFIX_UIO_MOBILE)
+  }
+
+  if (key.startsWith('bpo') && process.env.URL_PREFIX_BPO) {
+    return urls[key].replace('portal.edd.ca.gov', process.env.URL_PREFIX_BPO)
+  }
```
rocketnova, line 29: I think hard coding the default url prefixes here is less ideal. Did you consider making multiple env vars? For example:
```
uio-home-url-prefix: https://uio.edd.ca.gov/UIO/
uio-home-url-suffix: Pages/Public/ExternalUser/UIOnlineLandingPage.aspx
```
Then, this function would be concatenating the correct prefix with the correct suffix rather than doing a replace

components/TimeoutModal.tsx
```diff
@@ -55,7 +55,12 @@ export const TimeoutModal: React.FC<TimeoutModalProps> = ({ timedOut, userArrive
     // And at the end, send back to EDD
     setTimeout(() => {
       if (typeof window !== 'undefined') {
-        const eddLoginLink = getUrl('edd-log-in')?.concat(encodeURIComponent(window.location.toString()))
+        // Note that the concatenated portion of this link is functionally useless, as IDM is not currently
+        // able to redirect based on the resource_url parameter concatenated.
```
rocketnova, line 59: Why are we including the concatenated portion of the link if it's functionally useless?

components/TimeoutModal.tsx
```diff
@@ -55,7 +55,12 @@ export const TimeoutModal: React.FC<TimeoutModalProps> = ({ timedOut, userArrive
...
+        const eddLoginLink = getUrl('bpo-log-in')?.concat(
+          '?resource_url=',
```
rocketnova, line 61: I assume this addition is intentional? I see it is being added in this PR and want to double check

components/Header.tsx
```diff
@@ -57,7 +57,7 @@ export const Header: React.FC<HeaderProps> = ({ userArrivedFromUioMobile = false
           <Navbar.Toggle aria-controls="basic-navbar-nav" />
           <Navbar.Collapse>
             <Nav>
-              <Nav.Link target="_blank" rel="noopener noreferrer" href={uioHomeLink}>
+              <Nav.Link rel="noopener noreferrer" href={uioHomeLink}>
```
rocketnova, line 60: I assume that it's intentional that the link to UIO Home here _should not_ open in a new tab, but that the links above _should_?

README.md
```diff
@@ -26,7 +26,7 @@ relates to weekly certification
 - CERTIFICATE_DIR: The path to the client certificate (certificate must be in PFX/P12 format)
 - PFX_FILE: The name of the client certificate file
 - (Optional) PFX_PASSPHRASE: The import passphrase for the client certificate if there is one
-- (Optional) URL_UIO_LANDING, URL_UIOMOBILE_LANDING: The environment-specific links to the UIO and UIO Mobile landing pages
+- (Optional) URL_PREFIX_UIO, URL_PREFIX_UIO_MOBILE, URL_PREFIX_BPO: Environment-specific path prefixes for UIO and BPO links
```
rocketnova, line 29: Nit: I think it would clearer to break these out into individual lines

rocketnova (approved):
Approved, though I have some non-blocking suggestions and questions. Since this is time sensitive, I would also be happy for us to open a new ticket to revisit this at a future date.

components/TimeoutModal.tsx
kalvinwang: Because we didn't know it was functionally useless at the time, I imagine. I don't want to risk taking it out in this PR, so I added a comment instead.

components/TimeoutModal.tsx
kalvinwang: Yes, it was added here and removed from the URL in urls.json

components/Header.tsx
kalvinwang: Yes

components/TimeoutModal.tsx
rocketnova: 👍 Got it. Sounds good

kalvinwang:
Good suggestions, I'll merge now so I can release and test for EDD, but I'll create a ticket to follow up on these small things

kalvinwang:
Ticket created in https://github.com/cagov/ui-claim-tracker/issues/401

README.md
kalvinwang: done in a later pr

PR #402, opened 2021-08-21
Configure applicationinsights as a nodejs preloaded module

### LINKED ISSUE #367: Configure Application Insights with all the features we need
### Description

PR #288 added code to connect the Claim Tracker app directly to Azure Application Insights. But we've discovered that it also disabled Live Metrics and `requests`. We have temporarily reverted 288 with PR #366.

### Acceptance Criteria

- [ ] Re-enable application insights
- [ ] Applications Insights should capture `traces`, `requests`, and Live Metrics
- [ ] Check for other Application Insights features that should also be enabled

rocketnova:
## Ticket

- Resolves #367 

## Changes

- Configures the Azure Application Insights SDK using nodejs preloading
- Preloads `server-preload.js` on `yarn start`

## Context

The App Insights SDK needs to be called as early as possible in order for it to be able to automatically log HTTP requests. This approach allows us to avoid using a custom server, but still be able to leverage App Insights logging for Live Metrics and HTTP requests (i.e. `requests`).

## Testing

I've deployed this branch to the Internal Azure environment so that we can verify that it's working as expected.

1. Make a few requests against the site deployed on Internal, using a few different unique numbers
2. Run `traces | sort by timestamp desc` and `requests | sort by timestamp desc` in App Insights Logs for Internal against the last 30m and verify you can view your requests.

Note: This PR does not update pino to log to App Insights, so you won't see our custom `logger.error()` and `logger.info()` calls in App Insights yet.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@
 __mocks__
 tests/__snapshots__
 coverage
+server-preload.js
--- a/server-preload.js
+++ b/server-preload.js
@@ -1,0 +1,26 @@
+/**
+ * Use node.js preloading to require a node module before importing and executing
+ * the entrypoint script.
+ *
+ * Intended to be used in package.json scripts with something like:
+ * - `node -r ./server-preload.js ./node_modules/.bin/next start -p 8080`
+ *
+ * We need to connect to App Insights early enough for it to be able to log HTTP requests.
+ *
+ * References:
+ * - https://jake.tl/notes/2021-04-04-nextjs-preload-hack
+ * - https://nodejs.org/api/cli.html#cli_r_require_module
+ */
+
+function setupLogging() {
+  // Connect to application insights during preload to capture HTTP requests.
+  const appInsights = require('applicationinsights')
+  appInsights
+    .setup(process.env.APPLICATIONINSIGHTS_CONNECTION_STRING)
+    .setAutoCollectRequests(true) // Default setting, but explicitly set to true for clarity
+    .setSendLiveMetrics(true) // Send Live Metrics to App Insights
+    .setAutoCollectConsole(true, true) // Pass the second argument to also log console.log() statements
+    .start()
+}
+
+setupLogging()
--- a/server-preload.js
+++ b/server-preload.js
@@ -1,11 +1,11 @@
 /**
- * Use node.js preloading to require a node module before importing and executing
+ * We need to connect to App Insights early enough for it to be able to log HTTP requests,
+ * so we use node.js preloading to require a node module before importing and executing
  * the entrypoint script.
  *
- * Intended to be used in package.json scripts with something like:
+ * This script is called in package.json with:
  * - `node -r ./server-preload.js ./node_modules/.bin/next start -p 8080`
  *
- * We need to connect to App Insights early enough for it to be able to log HTTP requests.
  *
  * References:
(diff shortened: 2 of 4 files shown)


.eslintignore
```diff
@@ -3,3 +3,4 @@
 __mocks__
 tests/__snapshots__
 coverage
+server-preload.js
```
rocketnova, line 6: Note: I'm telling eslint to [ignore](https://eslint.org/docs/user-guide/configuring/ignoring-code) this file. Otherwise, it throws the following error:

```
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: server-preload.js.
The file must be included in at least one of the projects provided
```

server-preload.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+function setupLogging() {
+  // Connect to application insights during preload to capture HTTP requests.
+  const appInsights = require('applicationinsights')
+  appInsights
+    .setup(process.env.APPLICATIONINSIGHTS_CONNECTION_STRING)
+    .setAutoCollectRequests(true) // Default setting, but explicitly set to true for clarity
+    .setSendLiveMetrics(true) // Send Live Metrics to App Insights
+    .setAutoCollectConsole(true, true) // Pass the second argument to also log console.log() statements
```
rocketnova, line 22: I'm not 💯 on whether or not we want to also log Next.js's `console.log()` statements. It seemed useful (you can see them in `traces`, but we can easily turn this off.

kalvinwang: I think we shouldn't/can't because we'll end up logging claims data (which is not tied to any identifier, so perhaps not an actual issue, but potentially a perceived one)

rocketnova: I think the intention is to disable all `console.log` and `console.dir` calls, so I don't think that should be an issue.

server-preload.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Use node.js preloading to require a node module before importing and executing
+ * the entrypoint script.
+ *
+ * Intended to be used in package.json scripts with something like:
+ * - `node -r ./server-preload.js ./node_modules/.bin/next start -p 8080`
+ *
+ * We need to connect to App Insights early enough for it to be able to log HTTP requests.
```
kalvinwang, line 8: Minor edit for clarity
```suggestion
 * We need to connect to App Insights early enough for it to be able to log HTTP requests,
 * so we use node.js preloading to require a node module before importing and executing
 * the entrypoint script.
 *
 * This script is called in package.json with:
 * - `node -r ./server-preload.js ./node_modules/.bin/next start -p 8080`
 *
```

server-preload.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * References:
```
kalvinwang, line 10: Nice thanks!

kalvinwang (approved):
Lgtm, seems like a great alternative solution, thanks for finding and implementing it!

server-preload.js
kalvinwang: That's true, but I'm thinking that if we really want to log something, it's not too difficult for us to use the logger instead of console.log, whereas there is a nonzero chance that we will accidentally end up deploying a console.log to production at some point, because we are human :P

rocketnova: Let's talk more about this synchronously. `console.log()` logs still get logged to Azure.

PR #407, opened 2021-08-23
Refactor logging into a shared utility

### LINKED ISSUE #320: Refactor logging into separate module and support additional pino features
### Description

Our pino logger is currently instantiated in `index.tsx`, which makes it hard to call in helper util functions. Instead, let's support something like this:

```js
// imports a pino logger instance of `require('pino')()`
const parentLogger = require('./lib/logger')
const log = parentLogger.child({module: 'foo'})
```

### Acceptance Criteria

- [ ] Move logger functionality into separate module
- [ ] Add support for [exit logging](https://github.com/pinojs/pino/blob/master/docs/help.md#exit-logging)
- [ ] Add support for [child loggers](https://github.com/pinojs/pino/blob/master/docs/child-loggers.md)
- [ ] Replace logging-related `@TODO:`s with calling the logging module

### LINKED ISSUE #332: Combine logger calls about incoming request into one logger call
### Description

#330 introduces a second logger.info call about the incoming request: 

```
logger.info(req)
logger.info(query)
```

Ideally, these should be combined into one logger call.

### Acceptance Criteria

- [ ] All information about the incoming request should be logged in one call

rocketnova:
## Ticket

- Resolves #332 
- Resolves #320 

## Changes

- Re-adds all functionality from #288 that was temporarily reverted in #366 (and that was not included in #402), including:
  - Adds [`pino-applicationinsights`](https://github.com/ovhemert/pino-applicationinsights) as a dependency to log directly to Azure Application Insights in Production
  - Adds a type declaration file for `pino-applicationinsights`
- Creates a shared logger utility that can be called in the back end code
- Combines logging for `request` and `query` data into one call

## Context

This PR adds our logger shared utility to log directly to Azure Application Insights.

## Testing

I've deployed this branch to the Internal Azure environment so that we can verify that it's working as expected.

1. Make a few requests against the site deployed on Internal, using a few different unique numbers
2. Run `traces | sort by timestamp desc` in App Insights Logs for Internal against the last 30m and verify you can view data about your request. Right now, for each request to the app, you should see:
  - a `Request` log entry with request and query data (you should see the unique number you sent in the header)
  - a `ClaimData` log entry with the returned claim data from the API gateway
  - a `ScenarioContent` log entry with the FE props
  - a `Program Type` log entry that lists which program type is returned
3. Run `requests | sort by timestamp desc` to make sure behavior from #402 has not reverted.

Note: Right now, we have redundant `console.log` calls for claim data. Please ignore those.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,11 +1,11 @@
 import Head from 'next/head'
 import Container from 'react-bootstrap/Container'
-import pino from 'pino'
 import { ReactElement } from 'react'
 import { useTranslation } from 'next-i18next'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import { GetServerSideProps } from 'next'
 import Error from 'next/error'
+import pino from 'pino'
 
 import { Header } from '../components/Header'
 import { Title } from '../components/Title'
@@ -14,9 +14,10 @@ import { ClaimSection } from '../components/ClaimSection'
 import { TimeoutModal } from '../components/TimeoutModal'
 import { Footer } from '../components/Footer'
 
-import queryApiGateway, { getUniqueNumber } from '../utils/queryApiGateway'
-import getScenarioContent from '../utils/getScenarioContent'
 import { ScenarioContent } from '../types/common'
+import getScenarioContent from '../utils/getScenarioContent'
+import { Logger } from '../utils/logger'
+import queryApiGateway, { getUniqueNumber } from '../utils/queryApiGateway'
 
 export interface HomeProps {
   scenarioContent: ScenarioContent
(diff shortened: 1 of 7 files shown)


pages/index.tsx
```diff
@@ -85,44 +87,70 @@ export default function Home({
...
+        pino.error(error, 'Application error')
+        errorCode = 500
+      }
+    }
   }
 
-  // Note whether the user came from the main UIO website or UIO Mobile, and match
-  // that in our links back out to UIO.
-  const userArrivedFromUioMobile = query?.from === 'uiom'
-
```
rocketnova, line 125: Note: I've moved this to the top of the function.

PR #415, opened 2021-08-24
Update environment-specific links to work in timeout modal

kalvinwang:
#400 added support for environment-specific links; however, it didn't work for links in the timeout modal, because the timeout modal generates the URL on the client side, and environment variables can't be accessed client-side. This PR imports the link env vars on the server side in index.tsx and makes them available to be passed into getUrl which generates the environment-specific links.

Tested locally and in the Azure test env. Change `WARNING_TIMER * ONE_MINUTE_MS` to 10000 (milliseconds) and `REDIRECT_TIMER * ONE_MINUTE_MS` to 20000 in TimeoutModal.tsx to make testing quicker.

===

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,10 @@ relates to weekly certification
 - CERTIFICATE_DIR: The path to the client certificate (certificate must be in PFX/P12 format)
 - PFX_FILE: The name of the client certificate file
 - (Optional) PFX_PASSPHRASE: The import passphrase for the client certificate if there is one
-- (Optional) URL_PREFIX_UIO, URL_PREFIX_UIO_MOBILE, URL_PREFIX_BPO: Environment-specific path prefixes for UIO and BPO links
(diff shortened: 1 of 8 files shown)


rocketnova (approved):
This looks good to me as is, but a few suggestions/reminders:

- Don't forget to update the Environment Variables section in the Foundations doc with the new env vars
- It might make sense to allow overriding the timeout modal using env vars

rocketnova:
Oh one additional thought I forgot to add earlier. I'm a little anxious that this PR does not contain any tests. If we don't have time to add tests due to time sensitivity, can we open a tech debt ticket for that?

kalvinwang:
@rocketnova yes, agree with everything you wrote! i almost added the env var overriding for the timeout modal but figured it wasn't the best use of time right now :P will make a new ticket for tests too

PR #442, opened 2021-09-01
Throws an error on mismatched Unique Number

### LINKED ISSUE #386: If the API gateway response does not contain the same unique number we sent, we should display an error
### Description

The API gateway response includes a `uniqueNumber` property. We should verify that the response contains the same unique number that we requested. If the value does not match what we sent or if the value is `null`, then we should log the error and display a 500 response.

### Acceptance Criteria

- [ ] Mismatched unique number in API gateway response is logged and returns 500
- [ ] Null unique number in API gateway response is logged (different log message) and returns 500

lomky:
Checks for different Unique Numbers in the header & the API response.

Also updates our other errors in the query API file to re-throw their errors.

Test via `yarn test`

===

Resolves #386

- [ ] Mismatched unique number in API gateway response is logged and returns 500
- [ ] Null unique number in API gateway response is logged (different log message) and returns 500


tests/utils/queryApiGateway.test.ts
```diff
@@ -31,7 +31,7 @@ const goodRequest = {
 
 // Test queryApiGateway()
 describe('Querying the API Gateway', () => {
-  const goodResponse = { ClaimType: 'PUA' }
```
rocketnova, line 34: Note: This is using the old data structure, so I updated it to use current data structure.

tests/utils/queryApiGateway.test.ts
```diff
@@ -71,7 +71,7 @@ describe('Querying the API Gateway', () => {
       API_URL: goodUrl,
     })
 
-    const data = await queryApiGateway(goodRequest)
+    const data = await queryApiGateway(goodRequest, goodUniqueNumber)
```
rocketnova, line 74: Note: For some weird reason, linting wasn't erroring on the fact thaht `queryApiGateway()` requires a second argument. Fixed in this PR.

tests/utils/queryApiGateway.test.ts
```diff
@@ -81,18 +81,18 @@ describe('Querying the API Gateway', () => {
...
+    fetch.mockRejectedValue(new Error(networkErrorMessage))
     /* eslint-enable  @typescript-eslint/no-unsafe-call */
 
     // Mock process.env
     const restore = mockEnv({
       API_URL: goodUrl,
     })
 
-    const data = await queryApiGateway(goodRequest)
+    await expect(queryApiGateway(goodRequest, goodUniqueNumber)).rejects.toThrow(networkErrorMessage)
```
rocketnova, line 94: @lomky This is the [correct syntax](https://jestjs.io/docs/asynchronous#resolves--rejects) for testing async failures in jest.

tests/utils/queryApiGateway.test.ts
```diff
@@ -107,17 +107,16 @@ describe('Querying the API Gateway', () => {
       status: 403,
     })
     /* eslint-enable  @typescript-eslint/no-unsafe-assignment */
-    fetch.mockRejectedValue(errorResponse403)
+    fetch.mockResolvedValue(errorResponse403)
```
rocketnova, line 110: Note: I don't have a great grasp of `fetch`, but I believe the issue here is that a 403 response isn't actually a rejected promise. It's a resolved promise that contains the 403 status code.

utils/queryApiGateway.ts
```diff
@@ -176,6 +176,16 @@ export default async function queryApiGateway(req: IncomingMessage, uniqueNumber
...
+    throw error
+  }
+
+  if (apiData?.uniqueNumber !== uniqueNumber) {
+    const mismatchError = new Error(
+      `Mismatched API response and Header unique number (${apiData.uniqueNumber || 'null'} and ${uniqueNumber})`,
+    )
+    const logger: Logger = Logger.getInstance()
+    logger.log('error', mismatchError, 'Unexpected API gateway response')
+    throw mismatchError
```
rocketnova, line 188: @lomky I hope you don't mind: I did a little refactor of your code here because we don't need to do a `try catch` here.

tests/utils/queryApiGateway.test.ts
lomky: that makes sense to me!

tests/utils/queryApiGateway.test.ts
lomky: #444

utils/queryApiGateway.ts
```diff
@@ -176,6 +176,16 @@ export default async function queryApiGateway(req: IncomingMessage, uniqueNumber
...
     const logger: Logger = Logger.getInstance()
     logger.log('error', error, 'API gateway error')
+    throw error
+  }
+
+  if (apiData?.uniqueNumber !== uniqueNumber) {
+    const mismatchError = new Error(
+      `Mismatched API response and Header unique number (${apiData.uniqueNumber || 'null'} and ${uniqueNumber})`,
+    )
+    const logger: Logger = Logger.getInstance()
```
kalvinwang, line 186: just curious, why do we declare new instances of Logger each time we call logger.log in this file / would it be bad to just declare it once in the outer scope?

tests/utils/queryApiGateway.test.ts
```diff
@@ -161,16 +162,51 @@ describe('Querying the API Gateway', () => {
...
+    )
+
+    expect(fetch).toHaveBeenCalledTimes(1)
+    expect(loggerSpy).toHaveBeenCalledWith('error', expect.anything(), 'Unexpected API gateway response')
+
+    // Restore env vars
+    restore()
+  })
+
+  it('handles null unique number responses', async () => {
```
kalvinwang, line 192: I'm missing something... on line 199, we're sending null as the unique number? If the unique number is null, shouldn't we not be sending that request at all? And if we do send null, wouldn't it actually be matched (we send uniqueNumber: null, we get back uniqueNumber: null, `apiData?.uniqueNumber === uniqueNumber`)?

kalvinwang (commented):
Code lgtm but I'm confused by one of the tests!

tests/utils/queryApiGateway.test.ts
lomky: this is if we send a number and get back null, not if we send null!

lomky: does this clarify?
```suggestion
  it('handles API response with a null unique number', async () => {
```

utils/queryApiGateway.ts
lomky: it handles that! Rocket implemented the logger as a singleton, so `getInstance` will get a new one only if it hasnt aready been invoked, otherwise it retrieves the existing one

https://github.com/cagov/ui-claim-tracker/blob/ed59fd1e2b33cafc41fb709143f457e39ac012fb/utils/logger.ts#L35-L43

tests/utils/queryApiGateway.test.ts
rocketnova: Thanks @lomky !

utils/queryApiGateway.ts
kalvinwang: Oh cool, makes sense! Still curious if it would be bad (whether for readability or functionality) to just declare it once in the outer scope, maybe in a beforeAll or something. Maybe also a question for @rocketnova

kalvinwang (approved):
lgtm, thanks!

utils/queryApiGateway.ts
rocketnova: @kalvinwang  👀 What is a `beforeAll` in this context? 

There isn't really any harm in calling `const logger: Logger = Logger.getInstance()` at the top of `queryApiGateway()` but since we only need it in three edge cases, I figured it was minimal enough to only call as necessary.

kalvinwang: gotcha ok that makes sense, just checking! in jest, beforeAll is like beforeEach, but it's only run once instead of each time

lomky: we're in a utility file though, not a test file

PR #443, opened 2021-09-01
Displays None when no last payment provided

### LINKED ISSUE #419: Display "None" if lastPaymentIssued or lastPaymentAmount returns null
### Description

This ticket updates behavior that was implemented in #365 related to the Last Payment Issued field. We have since learned that in UIO, this field displays "None" when null, and the requirement is to match that functionality in the Claim Tracker.

In the Claim Details display, "Last Payment Issued" combines data from `lastPaymentIssued` and `lastPaymentAmount`. If either field returns null, we should display the field value as "None" for "Last Payment Issued". 

Currently, we hide the display field when either API field returns null, so this ticket is to update the behavior to display "None" under these conditions instead of hiding the field. 

### Acceptance Criteria

- [ ] When either lastPaymentIssued of lastPaymentAmount returns null, the field value for Last Payment Issue under Claim Details displays as "None"

lomky:
What it says on the tin!

Test in storybook:
- Component level, use storybook to manually test
  - Set the value to empty on the component http://localhost:6006/?path=/story/component-page-section-claim-details--claim-details
- API level & component level are also both tested in the tests
  - `yarn test` 

===

Resolves #419

- [x]  When either lastPaymentIssued of lastPaymentAmount returns null, the field value for Last Payment Issue under Claim Details displays as "None"

### DIFF
--- a/components/ClaimDetails.tsx
+++ b/components/ClaimDetails.tsx
@@ -29,13 +29,6 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
     )
   }
 
-  let lastPaymentIssuedField: JSX.Element | null = null
-  if (lastPaymentIssued) {
-    lastPaymentIssuedField = (
-      <InfoField loading={loading} label={t('claim-details.last-payment-issued')} text={lastPaymentIssued} />
-    )
-  }
-
   let extensionField: JSX.Element | null = null
   if (extensionType) {
     extensionField = (
@@ -61,7 +54,11 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
 
             {weeklyBenefitAmountField}
 
-            {lastPaymentIssuedField}
+            <InfoField
+              loading={loading}
+              label={t('claim-details.last-payment-issued')}
+              text={lastPaymentIssued || t('claim-details.none')}
+            />
           </div>
           {extensionField}
         </div>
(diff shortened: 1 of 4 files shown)


components/ClaimDetails.tsx
```diff
@@ -61,7 +54,11 @@ export const ClaimDetails: React.FC<ClaimDetailsProps> = ({
 
             {weeklyBenefitAmountField}
 
-            {lastPaymentIssuedField}
+            <InfoField
+              loading={loading}
+              label={t('claim-details.last-payment-issued')}
+              text={lastPaymentIssued || t('claim-details.none')}
```
lomky, line 60: This is built on top of existing tested logic that sets this field to null in either case.

Logic:
https://github.com/cagov/ui-claim-tracker/blob/d0ac954ace870af5b203f2708c5bc6da10bda166/utils/getClaimDetails.ts#L176-L179

Tests:
https://github.com/cagov/ui-claim-tracker/blob/d0ac954ace870af5b203f2708c5bc6da10bda166/tests/utils/getClaimDetails.test.ts#L176-L226

lomky: also we set it here instead of back there because _this_ is the layer of our code where we resolve our translations

rocketnova (approved):
🚢 

A couple thoughts:
- You've got a test failing
- For completeness, we could add a flag to `apiGatewayStub.ts` to set `lastPaymentIssued` or `lastPaymentAmount` to `''`, but not required

lomky:
see this is what happens when you try to finish a ticket at EOD, you forget to commit the tests!

PR #445, opened 2021-09-02
Strip trailing slash from API_URL

### LINKED ISSUE #438: Strip trailing slashes from environment variable URLs
### Description

The presence or absence of a trailing slash on an environment variable that contains a URL shouldn't affect anything, but it does, at least for API_URL. (The endpoint needs to be `path?x=y` not `path/?x=y`). We can update our code to strip trailing slashes in case one is accidentally added in App Service App Settings again. 

Edit: moved the task for removing trailing slashes from other env vars to #424, to reduce the risk of this change

### Acceptance Criteria

- [ ] Remove trailing slashes that might appear in environment variables

kalvinwang:
This PR strips the trailing slash from the API_URL environment variable, since the endpoint in the URL should never contain a trailing slash. 

===

Resolves #438

### DIFF
--- a/utils/queryApiGateway.ts
+++ b/utils/queryApiGateway.ts
@@ -50,7 +50,8 @@ export function getApiVars(): ApiEnvVars {
   apiEnvVars.idHeaderName = process.env.ID_HEADER_NAME ?? ''
 
   // API fields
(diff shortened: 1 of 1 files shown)


rocketnova (commented):
LGTM! Thanks!

PR #451, opened 2021-09-03
Include Google Analytics code

### LINKED ISSUE #313: Add Google Analytics tag
### Description

Add Google Analytics tag, on production only

Send analytics to both EDD and ODI's GA accounts like we did last year

### Acceptance Criteria

- [ ] Confirm Google Analytics is working on production (use the manual header method to test)

kalvinwang:
This PR adds the Google Analytics code snippet. It also creates a new environment variable, ENABLE_GOOGLE_ANALYTICS, that must be set to 'enabled' for the GA snippet to render.

* Tested by going to Google Analytics Realtime, searching for /claimstatus, and seeing results 
* Tested environment variable locally to see if network requests appear in each case

===

Resolves #313

- [x] Relevant documentation (e.g. READMEs, Technical Foundation) updated

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -25,6 +25,7 @@ export interface HomeProps {
   errorCode?: number | null
   userArrivedFromUioMobile?: boolean
   urlPrefixes: UrlPrefixes
+  enableGoogleAnalytics: string
 }
 
 export default function Home({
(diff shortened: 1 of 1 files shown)


lomky (approved):

pages/index.tsx
```diff
@@ -75,6 +101,7 @@ export default function Home({
           href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400;1,700&display=swap"
           rel="stylesheet"
         />
+        {enableGoogleAnalytics === 'enabled' && googleAnalytics}
```
rocketnova, line 104: I think I might be being dense here. Doesn't this expression evaluate to `true`/`false`?

pages/index.tsx
```diff
@@ -99,6 +126,9 @@ export const getServerSideProps: GetServerSideProps = async ({ req, res, locale,
     urlPrefixBpo: process.env.URL_PREFIX_BPO ?? '',
   }
 
+  // Set to 'enabled' to include Google Analytics code
+  const ENABLE_GOOGLE_ANALYTICS = process.env.ENABLE_GOOGLE_ANALYTICS ?? ''
```
rocketnova, line 130: Do we need to be alerted on whether this env var is missing? If so, there is code in `queryApiGateway.ts` that could be re-used or generalized.

pages/index.tsx
```diff
@@ -34,9 +35,34 @@ export default function Home({
...
+        gtag('js', new Date())
+        // For details see: https://support.google.com/analytics/answer/9310895?hl=en
+        // https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization
+        gtag('config', 'UA-3419582-2', { 'anonymize_ip': true }) // www.ca.gov
+        gtag('config', 'UA-3419582-31', { 'anonymize_ip': true }) // edd.ca.gov`
+
+  const googleAnalytics = (
+    <>
+      {/* Global site tag (gtag.js) - Google Analytics */}
+      <script async src="https://www.googletagmanager.com/gtag/js?id=UA-3419582-2" />
```
rocketnova, line 57: I feel like there's a teeny bit of a bad code smell here with the hardcoded IDs.

pages/index.tsx
kalvinwang: It's a JSX feature, react conditional rendering: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

pages/index.tsx
kalvinwang: Oh interesting! I guess ideally we could be alerted if it's missing in production, and only in production. I don't think we currently have a way of detecting whether we're in production (although it's easy enough to add an env var for that), so I'll create a ticket to alert if it's missing in prod, how's that sound?

pages/index.tsx
rocketnova: 😖

pages/index.tsx
rocketnova: We do! We're doing that for the missing unique number.

pages/index.tsx
kalvinwang: Hmm. Is there a constants file I could move this to? It doesn't make sense for it to be an env var.

pages/index.tsx
lomky: oh wow Javascript doesn't do this natively? TIL

pages/index.tsx
kalvinwang: (talked to Rocket, we don't have a way *in our codebase* of detecting whether we're in prod)

created: https://github.com/cagov/ui-claim-tracker/issues/460

Issue #452, opened 2021-09-03
Check in with UIB about system page content and translation needs


nicoleslaw:
### Description

We have ~3 system pages and want to run these by UIB and WCUG for input. This is not launch-blocking as it would be an enhancement over the baseline UIO experience to have translated system messages.

### Acceptance Criteria

- [ ] Share screenshot and existing content with UIB
- [ ] Get input on content and translation needs
- [ ] Implement any changes in a PR if possible without eng support

nicoleslaw:
Closing in favor of #467

PR #453, opened 2021-09-03
Adds custom error pages

### LINKED ISSUE #123: Update default Next.js error pages

lomky:
Error Pages! Compare to the designs on the "CA UI - Claim Status" figma, tab "SP11 - 7/21" 

Status Code | Image
--|--
404 | Claim Status Tracker - 404 error page
All others (500s) | Claim Status Tracker - 500 error page

 
Testing
 - both pages can be previewed in Storybook (under Page > Errored and 404)
 - navigating to http://localhost:3000/ will show the 404
 - navigating to http://localhost:3000/claimstatus with no unique number will show the 500

===

Resolves #123 

- [x] Show the appropriate error page (404 or 5xx) when triggered

### DIFF
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -52,6 +52,6 @@ export default function Custom404({ userArrivedFromUioMobile = false }: Custom40
 
 export const getStaticProps: GetStaticProps = async ({ locale }) => ({
   props: {
-    ...(await serverSideTranslations(locale, ['common'])),
+    ...(await serverSideTranslations(locale || 'en', ['common'])),
   },
 })
--- a/pages/404.tsx
+++ b/pages/404.tsx
@@ -1,0 +1,55 @@
+import Head from 'next/head'
+import { useTranslation } from 'next-i18next'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { ReactElement } from 'react'
+import Container from 'react-bootstrap/Container'
+
+import { Footer } from '../components/Footer'
+import { Header } from '../components/Header'
+import { Button } from '../components/Button'
+import { UrlPrefixes } from '../types/common'
+import getUrl from '../utils/browser/getUrl'
+
+export interface Custom404Props {
+  userArrivedFromUioMobile?: boolean
+}
+
+export default function Custom404({ userArrivedFromUioMobile = false }: Custom404Props): ReactElement {
+  const { t } = useTranslation('common')
+
+  function redirectToUIHome() {
+    const uioHomeLink = userArrivedFromUioMobile ? getUrl('uio-mobile-home-url') : getUrl('uio-desktop-home-url')
+    window.location.href = uioHomeLink || ''
+  }
+
+  return (
+    <Container fluid className="index">
+      <Head>
+        <title>{t('title')}</title>
+        <link rel="icon" href="/claimstatus/favicon.ico" />
+        <link rel="preconnect" href="https://fonts.googleapis.com" />
+        <link rel="preconnect" href="https://fonts.gstatic.com" crossOrigin="true" />
+        <link
+          href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400;1,700&display=swap"
+          rel="stylesheet"
+        />
+      </Head>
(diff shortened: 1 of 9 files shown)


pages/404.tsx
```diff
@@ -0,0 +1,57 @@
+import { GetStaticProps } from 'next'
```
lomky, line 1: > Kat, a lot of this feels pretty redundant to the index page - whats happening here?

SO Next.js has a strong opinion that the 404 page should be rendered at build time, which is reasonable! its always identical. But our header & footer are loaded on the main page server side. This is annoying in redundancy and I bet we can extract it out somehow, but I think this redundancy is the least-evil option for go-live

lomky: See https://nextjs.org/docs/advanced-features/custom-error-page#404-page

pages/_error.tsx
```diff
@@ -0,0 +1,34 @@
+import { NextPageContext } from 'next'
```
lomky, line 1: So unlike the 404 page Next offers both [a static 500](https://nextjs.org/docs/advanced-features/custom-error-page#customizing-the-500-page) option and [a more flexible one](https://nextjs.org/docs/advanced-features/custom-error-page#more-advanced-error-page-customizing). We're going with the second one and clobbering it a little bit, returning the same component regardless of what the status code is.

public/locales/es/common.json
```diff
@@ -9,7 +9,17 @@
...
+    "not-found": {
+      "label": "We can't find that page.",
+      "entry": "The link you clicked may be broken or the page may have been removed."
+    },
+    "button": "Return to UI Home",
+    "server-error": {
+      "label": "There’s an error on our end.",
+      "entry": "We are unable to perform your request."
+    }
+  },
```
lomky, line 22: As discussed in the slack, we're not translating these yet

pages/404.tsx
rocketnova: Sounds good for launch. I've opened a low priority reminder ticket about this: https://github.com/cagov/ui-claim-tracker/issues/457

pages/_error.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <div className="error-entry">{t('errors.server-error.entry')}</div>
+      <Button primary label={t('errors.button')} onClick={redirectToUIHome} />
+    </div>
+  )
+}
+
+Error.getInitialProps = ({ res, err }: NextPageContext) => {
+  const statusCode = res ? res.statusCode : err ? err.statusCode : 404
+  return { statusCode }
+}
```
rocketnova, line 32: I don't think we need this because we are not using the `statusCode`. Is that right?

public/locales/es/common.json
rocketnova: @nicoleslaw Do you want a ticket reminder for this?

rocketnova (approved):
LGTM!

public/locales/es/common.json
kalvinwang: Nicole already created one! https://github.com/cagov/ui-claim-tracker/issues/452

pages/_error.tsx
lomky: oooo you're right! nice

kalvinwang (approved):
lgtm, thanks!

Issue #376, opened 2021-08-12
Create maintenance page


kalvinwang:
### Description

Be able to enable a maintenance page by setting an environment variable

We’ll only manually enable the maintenance page / it will only be shown when a) something is seriously wrong, and b) EDD has not yet flipped their switch to hide the links to the claim tracker (edit: or claimants visit their bookmark of the claim tracker directly). So this page should be seen rarely, much less often than, say the custom 500 error page.

### Acceptance Criteria

- [ ] Test maintenance page on staging

nicoleslaw:
Here is a quick suggestion based on some UI Online content but I can follow up with UIB soon for any updates. cc @dianagriffin for input

—

## We’ll be back soon

This page is currently unavailable during scheduled system maintenance. Thank you for your patience as we continue to improve our services.

PR #494, opened 2021-09-17
Adds storybooks for Maintenance Component & Page under Maintenance

lomky:
Adds stories for the Maintenance compoenent itself, and for the whole Page in maintenance mode.

Tweaks the styling to properly render in Storybook.

===

Enhances #376

### DIFF
--- a/stories/Page.stories.tsx
+++ b/stories/Page.stories.tsx
@@ -44,7 +44,7 @@ export default {
     },
     errorCode: {
       control: {
-        type: 'text',
+        type: 'number',
       },
     },
   },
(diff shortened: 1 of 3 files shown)


stories/Page.stories.tsx
```diff
@@ -70,11 +70,14 @@ Default.args = {
   errorCode: null,
```
rocketnova, line 70: Nit: I think we should go ahead and pull out the `errorCode` control. It doesn't make as much sense anymore if we have a separate story for Errored and because we are no longer displaying the specific error code to the user.

rocketnova (approved):
LGTM!

stories/Page.stories.tsx
lomky: since errorCode is an actual HomeProp, this line just makes the type show up properly (a number instead of a JSON) rather than whether the arg shows up or not.

PR #501, opened 2021-09-22
Adds an error catch for null responses

### LINKED ISSUE #480: Properly log & alert on null response with a Unique Number
### Description

We have found a response that follows the `null` response formulation, but _does_ have a Unique Number in the response. 

Right now this `500`s with an error on attempting to parse the benefit year. Far better to catch this earlier in the process & log a null case. 

Not sure whether 'null' should be a Sev-3 or Sev-4.

### Acceptance Criteria

- [ ] `null` responses (with and without Unique Number) create a distinct log & alert

lomky:
Checks if the API has responded with what we consider a null response, and logs that case clearly.

===

Resolves #480 

- [ ] Manual Browser Testing performed (with modheader, either locally or on BrowserStack)

### DIFF
--- a/utils/queryApiGateway.ts
+++ b/utils/queryApiGateway.ts
@@ -14,6 +14,7 @@
 import path from 'path'
 import fs from 'fs'
 import https from 'https'
+import assert from 'assert'
 import { IncomingMessage } from 'http'
 import { Claim } from '../types/common'
 import { Logger } from './logger'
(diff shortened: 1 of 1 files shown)


utils/queryApiGateway.ts
```diff
@@ -112,6 +113,51 @@ export function getUniqueNumber(req: IncomingMessage): string {
   return req.headers[idHeaderName.toLowerCase()] as string
 }
 
+/**
+ * Check if the API responded with a null response body
+ *
+ * We do not expect any of these in production, in theory.
+ */
+export function reponseIsNull(apiBody: Claim): boolean {
+  const response: Claim = JSON.parse(JSON.stringify(apiBody)) as Claim
```
rocketnova, line 122: I think you should do this instead:
```suggestion
  const response = extractJSON(apiBody)
```
But you should do it inside a `try catch` block (either in this function or the function calling this function) because JSON parsing can throw errors

utils/queryApiGateway.ts
```diff
@@ -112,6 +113,51 @@ export function getUniqueNumber(req: IncomingMessage): string {
...
+  delete response.uniqueNumber
+
+  // This is the response for a bad UniqueNumber, or a UniqueNumber without a claim
+  const nullResponseShort = {
+    claimDetails: null,
+    hasCertificationWeeksAvailable: false,
+    hasPendingWeeks: false,
+    pendingDetermination: null,
+  }
```
rocketnova, line 131: If you include this, you are in effect making it so that this is unreachable: https://github.com/cagov/ui-claim-tracker/blob/d79ff60b15969589f2e9da3f7342d1d82ecbd1e6/utils/queryApiGateway.ts#L183-L190

utils/queryApiGateway.ts
```diff
@@ -112,6 +113,51 @@ export function getUniqueNumber(req: IncomingMessage): string {
...
+      monetaryStatus: '',
+    },
+    hasCertificationWeeksAvailable: false,
+    hasPendingWeeks: false,
+    pendingDetermination: [],
+  }
+
+  try {
+    assert.notDeepStrictEqual(response, nullResponseShort, 'Response is null')
+    assert.notDeepStrictEqual(response, nullResponseLong, 'Response is null')
```
rocketnova, line 153: This is neat

utils/queryApiGateway.ts
```diff
@@ -112,6 +113,51 @@ export function getUniqueNumber(req: IncomingMessage): string {
...
+      claimBalance: null,
+      weeklyBenefitAmount: null,
+      lastPaymentIssued: null,
+      lastPaymentAmount: null,
+      monetaryStatus: '',
+    },
+    hasCertificationWeeksAvailable: false,
+    hasPendingWeeks: false,
+    pendingDetermination: [],
+  }
```
rocketnova, line 149: I feel like hard coding these "null" responses is not the right way to approach this, but I think that #150 is actually the right way, so this seems like a good enough approach until we get #150 done.

utils/queryApiGateway.ts
rocketnova: Therefore, I think you should either not include the `nullResponseShort` or handle the mismatchError first before you handle the new null response error.

lomky: oh interesting - I was considering `mismatched UniqueNumber` when the API gave us a `null` to be incidentally catching the `null` response case, and that this is a more accurate error for that situation. That is still reachable if we get back the different unique number, or get back some claim data but a null unique number.

rocketnova: Oh interesting. I see what you mean. This sounds good to me!

utils/queryApiGateway.ts
lomky: (I will add a comment explaining this syntax)

lomky:
```suggestion
  // This is a trick to make a deep-copy of a JSON object
  const response: Claim = JSON.parse(JSON.stringify(apiBody)) as Claim
```

PR #515, opened 2021-10-12
Run CI on future and PRs against future.

lomky:
Make our CI work against all three long-lived branches.


rocketnova (approved):
Thank you!

PR #531, opened 2021-10-27
Adds FED-ED, Pandemic, & Other Federal Extensions BYE Split

### LINKED ISSUE #528: Add BYE Scenario breakout for Extention Scenarios
### Description

The final BYE Requirements include a detailed breakout for the Extensions into three scenarios instead of just one

### Acceptance Criteria

- [ ]

rocketnova:
## Ticket

- Resolves #528 

## Changes

- Re-orders scenarios 7-10
- Adds scenarios 11 & 12

## Context

We received additional content from EDD that also splits out a couple new scenarios.

## Testing

- [x] Test latest content doc for scenarios 7-12 against chromatic
- [x] Make sure the links display the Spanish version of the urls when the language selected is Spanish

### DIFF
--- a/utils/getScenarioContent.ts
+++ b/utils/getScenarioContent.ts
@@ -38,7 +38,7 @@ export const ScenarioTypeNames = {
   [ScenarioType.Scenario7]: 'Benefit year end: Regular UI',
   [ScenarioType.Scenario8]: 'Benefit year end: PUA',
   [ScenarioType.Scenario9]: 'Benefit year end: DUA',
-  [ScenarioType.Scenario10]: 'Benefit year end: Other Extensions',
+  [ScenarioType.Scenario10]: 'Benefit year end: Old Extensions',
   [ScenarioType.Scenario11]: 'Benefit year end: Pandemic Extensions',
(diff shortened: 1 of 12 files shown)


utils/getScenarioContent.ts
```diff
@@ -159,19 +163,32 @@ export function identifyPendingDeterminationScenario(
...
+    return true
+  }
+
+  return false
+}
+
+/**
+ * Identifies the Program Types that count as another extension
+ **/
+export function isOtherExtension(programType: string): boolean {
```
lomky, line 185: Given that we've pulled out all the distinct types until we only have the EU[CXYWZ] family left - do these have a proper category name instead of other?

lomky (approved):
One thought on `Other` extensions, otherwise approved!

utils/getScenarioContent.ts
lomky: `Regular`? `Normal`? Should we ask EDD?

rocketnova: We decided to use EDD's nomenclature and call these `Old Extensions`.

PR #538, opened 2021-11-03
Enforce Required API Fields

### LINKED ISSUE #524: Simplify business logic flow if isBye is always returned by the API
### Description

If it turns out that `isBye` is a garunteed field in the API response, we can simplify several pieces of logic, as noted by @rocketnova in #516. We will also need to update our 'null' error reponse catches to expect it.

This work is blocked until we see how the API handles the new field.

### Acceptance Criteria

- [ ] Simplify business logic if BYE is garunteed
- [ ] Update base API response expectations

lomky:
Removes the optional nature of the fields the API always returns, and simplifies some logic enabled by this.

All changes are a straight refactor - tests should all still pass, snapshots should be unaffected, testing in browser & storybook should not fail.

===

Resolves #524 

- [x] Simplify business logic if BYE is garunteed
- [x] Update base API response expectations


types/common.ts
```diff
@@ -37,24 +37,24 @@ export interface PendingDetermination {
...
-  claimBalance: number
-  weeklyBenefitAmount: number
-  lastPaymentIssued: string
-  lastPaymentAmount: number
+  benefitYearStartDate: string | null
+  benefitYearEndDate: string | null
+  claimBalance: number | null
+  weeklyBenefitAmount: number | null
+  lastPaymentIssued: string | null
+  lastPaymentAmount: number | null
```
lomky, line 45: These values are null in the long null response we sometimes see

utils/queryApiGateway.ts
```diff
@@ -126,7 +126,7 @@ export function reponseIsNullish(apiBody: Claim): boolean {
   // This is a trick to make a deep-copy of a JSON object
   const response: Claim = JSON.parse(JSON.stringify(apiBody)) as Claim
   const responseUniqueNumber = response.uniqueNumber
-  delete response.uniqueNumber
+  response.uniqueNumber = null
```
lomky, line 129: probably should have done it this way in the first place 🙈

types/common.ts
```diff
@@ -46,15 +46,35 @@ export interface ClaimDetailsResult {
...
+  uniqueNumber: null | string
+  claimDetails: null | ClaimDetailsResult
+  hasPendingWeeks: boolean // deprecated for hasValidPendingWeeks
+  hasValidPendingWeeks: boolean
+  hasCertificationWeeksAvailable: boolean
+  isBYE: boolean
+  pendingDetermination: null | PendingDetermination[]
+}
+
+export interface NullClaim {
```
lomky, line 69: By pulling `NullClaim` & `NullClaimDetailsResult` out into their own types, we can now narrow on them in `queryApiGateway` - TS then can tell when we return from `queryApiGateway` that we _must_ have `Claim`  instead of `NullClaim` - allowing us to assert not only that we expect certain fields exist, but their value in non-`null`, as well.

utils/queryApiGateway.ts
```diff
@@ -126,7 +126,7 @@ export function reponseIsNullish(apiBody: Claim): boolean {
   // This is a trick to make a deep-copy of a JSON object
   const response: Claim = JSON.parse(JSON.stringify(apiBody)) as Claim
```
lomky, line 127: without _some missing piece_ to avoid the `as`, we can't go further in the Typescript refactor. But we have met the goal of the initial ticket!

utils/queryApiGateway.ts
```diff
@@ -263,24 +278,23 @@ export default async function queryApiGateway(req: IncomingMessage, uniqueNumber
   }
```
rocketnova, line 278: After looking through the code, I actually think we need an additional first `if` statement. Something like: 

```typescript
// If apiData is falsey, throw an error
if (!apiData) {
  const nullError = new Error(`API responded with null (queried with ${uniqueNumber}, responded with ${
        apiData}`)
  logger.log(childLogger, 'error', nullError, 'Unexpected API gateway response')
  throw nullError
}
```

utils/queryApiGateway.ts
```diff
@@ -263,24 +278,23 @@ export default async function queryApiGateway(req: IncomingMessage, uniqueNumber
...
     )
     logger.log(childLogger, 'error', mismatchError, 'Unexpected API gateway response')
     throw mismatchError
   }
-
   // Yell if the API returns a null or null-ish response
-  if (reponseIsNullish(apiData)) {
+  else if (reponseIsNullish(apiData)) {
     const nullResponseError = new Error(
       `API responded with a null response (queried with ${uniqueNumber}, responded with ${
```
rocketnova, line 291:
```suggestion
      `API responded with a null object (queried with ${uniqueNumber}, returned unique number ${
```

tests/utils/queryApiGateway.test.ts
```diff
@@ -31,9 +31,13 @@ const goodRequest = {
 
 // Test queryApiGateway()
 describe('Querying the API Gateway', () => {
-  const goodResponse = {
+  const goodResponse: Claim = {
```
rocketnova, line 34: Let's call this something else since it's more like a nullish response. How about `mockedResponse`?

rocketnova (changes requested):
A few small changes. Otherwise, LGTM!

tests/utils/queryApiGateway.test.ts
```diff
@@ -180,6 +180,26 @@ describe('Querying the API Gateway', () => {
     restore()
   })
 
+  it('handles null fully null api response', async () => {
```
rocketnova, line 183:
```suggestion
  it('handles fully null api response', async () => {
```

rocketnova (approved):
One minor typo. LGTM!

PR #540, opened 2021-11-03
Updates BYE code to match API response

lomky:
Updates to match the API we are seeing in Dev!

===

- [ ] Test scenarios 7-12 against dev


rocketnova (approved):
LGTM! I tested against all the scenarios and the negative cases.

Note: I changed the base branch to `future`.

PR #541, opened 2021-11-03
Add updated BYE scenario logic for FED-ED

### LINKED ISSUE #539: Update BYE logic for Scenario 10 & Scenario Scenario 12
### Description

Scenario 10 logic should be updated from just EUC, EUX,  EUY, EUW, EUZ to:
- matching EUC, EUX,  EUY, EUW, EUZ
- OR FED-ED if `benefitYearEndDate` is before 05/10/2020

Scenario 12 logic should be updated from any FED-ED to:
- FED-ED only if `benefitYearEndDate` is on or after 05/10/2020

rocketnova:
## Ticket

- Resolves #539 

## Changes

- Adds logic to handle BYE scenarios for FED-ED program type based on benefitYearEndDate
- Adds a helper function for comparing date strings

## Context

Scenario 10 logic should be updated from just EUC, EUX,  EUY, EUW, EUZ to:
- matching EUC, EUX,  EUY, EUW, EUZ
- OR FED-ED if `benefitYearEndDate` is before 05/10/2020

Scenario 12 logic should be updated from any FED-ED to:
- FED-ED only if `benefitYearEndDate` is on or after 05/10/2020

Edge case is covered: If `isBYE` is true AND `programType` is `FED-ED Extension` AND `benefitYearEndDate` is javascript falsey (e.g. undefined, null, empty string, etc), then throw an error.

## Testing

- [x] Test against dev for a claim that has `isBYE=true`, `programType=FED-ED`, and `benefitYearEndDate<5/10/2020` == Scenario 10
- [x] Test against dev for a claim that has `isBYE=true`, `programType=FED-ED`, and `benefitYearEndDate>=5/10/2020` == Scenario 12
- [x] Test against latest https://github.com/rocketnova/ca-ui-mini-api-mock and use unique number `invalid_feded_bye` == error

### DIFF
--- a/utils/getScenarioContent.ts
+++ b/utils/getScenarioContent.ts
@@ -254,6 +254,8 @@ export function byeScenario(claimData: Claim): ScenarioType | null {
           } else {
             return ScenarioType.Scenario12
           }
+        } else {
+          throw new Error('Claim is marked as isBYE, but is missing benefit year end date value')
         }
     }
   }
--- a/utils/getScenarioContent.ts
+++ b/utils/getScenarioContent.ts
@@ -10,7 +10,7 @@
 import { Claim, ClaimDetailsContent, PendingDetermination, ScenarioContent } from '../types/common'
 import getClaimDetails from './getClaimDetails'
 import getClaimStatus from './getClaimStatus'
-import { isDatePast, isDateStringFalsy, isValidDate, parseApiGatewayDate } from './formatDate'
+import { isDatePast, isDateStringFalsy, isFirstDateBefore, isValidDate, parseApiGatewayDate } from './formatDate'
 import { isFirstTimeSlotEarlier } from './timeSlot'
 
 export enum ScenarioType {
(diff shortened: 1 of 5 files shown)


utils/getScenarioContent.ts
```diff
@@ -243,8 +244,17 @@ export function byeScenario(claimData: Claim): ScenarioType | null {
...
+      // on whether the benefitYearEndDate is before or on/after 2020-05-10.
       case 'FED-ED Extension':
-        return ScenarioType.Scenario12
+        if (benefitYearEndDate) {
+          if (isFirstDateBefore(benefitYearEndDate, '2020-05-10T00:00:00')) {
+            return ScenarioType.Scenario10
+          } else {
+            return ScenarioType.Scenario12
+          }
+        }
```
lomky, line 257: So in the (badly formatted API response) case that we have `isBYE` and `FED-ED` `programType`, but `benefitYearEndDate` is undefined - we will fall through to Scenario5 as we've written this. (we'll fall thru to return null, meaning 299 will be false, falling us through to scenario5 on line 305)

Do we think that's the best plan, or should we have one of 10/12 be the catch?

rocketnova: @lomky Yeah, I made that decision intentionally, but maybe tagging in @dianagriffin is the right next step here.

rocketnova: @lomky I have updated the code. I updated this PR description to include one more test case.

lomky (approved):
🎉 :shipit:

PR #557, opened 2021-11-22
Add tests for per-environment urls and standardize url naming convention

### LINKED ISSUE #424: Add tests/snapshots for environment-specific links
### Description

Add tests/snapshots for environment-specific links which were added in #400 and #415. 

Two additional small tasks, once the tests are added and it's possible to verify these don't change any output

1. As a small refactor, create the following prefixes in urls.json, and use them in getUrl.ts, so the strings-to-replace aren't located in getUrl.ts, and urls.json still only contains URLs. (Once that's deployed in each env, the URL_PREFIX_* env var values in each env will need to be updated from e.g. "uio.edd.ca.gov/UIO" to "https://uio.edd.ca.gov/UIO")

  "uio-prefix": "https://uio.edd.ca.gov/UIO",
  "bpo-prefix": "https://portal.edd.ca.gov/WebApp",
  "uio-mobile-prefix": "https://uiom.edd.ca.gov/UIOM",

2. To avoid potential bugs from unintentionally added trailing slashes, strip trailing slashes from all three environment variables, just like #438 did for API_URL. 

### Acceptance Criteria

- [ ] Deploy this to test and prod and verify that the links are still pointing to the appropriate environments

rocketnova:
## Ticket

- Resolves #424 

## Changes

- Adds tests for per-environment urls
- Strips trailing slashes from url prefix env vars
- Renames keys in `urls.json` to follow a consistent naming convention
- Renames env var `URL_PREFIX_STATUS` to `URL_PREFIX_UIO_CLAIMSTATUS` to follow a consistent naming convention
  - Note: This change will require updating env vars in all Azure environments on deploy.

## Context

This PR adds unit tests for per-environment urls and strips trailing slashes as specified in #424. However, I chose not to replace the use of `string.replace()` with a prefix/suffix structure in `urls.json`. In the end, I felt that the current structure in `urls.json` is easier to understand and keep updated. I felt that was more important than the minimal benefits we would get be refactoring.

## Testing

`yarn test`


tests/utils/getUrl.test.ts
```diff
@@ -10,3 +29,53 @@ describe('Retrieving urls', () => {
...
+    const mockedPrefix = 'mock-prefix'
+    const argPrefix = 'arg-prefix'
+
+    it(`returns the url with env var prefix when the env var is provided for ${envVar}`, () => {
+      const restore = mockEnv({
+        [envVar]: mockedPrefix,
+      })
+
+      const url = getUrl(linkKey)
+      expect(url).toEqual(expect.stringMatching(mockedPrefix))
```
kalvinwang, line 53: Shouldn't matter in practice but unless we expect argPrefix or mockedPrefix or toReplace to be regular expressions, I think it makes sense to use stringContaining instead of stringMatching everywhere in this section? [docs](https://jest-bot.github.io/jest/docs/expect.html#expectstringcontainingstring)

kalvinwang (approved):
Lgtm, thanks! Not doing the prefix/suffix refactor makes sense to me.

tests/utils/getUrl.test.ts
rocketnova: I don't think that's going to work because `stringContaining()` is for exact matches, whereas `stringMatching()` is for partial matches. From these [docs](https://jestjs.io/docs/expect#expectstringcontainingstring):

> expect.stringContaining(string) matches the received value if it is a string that contains the exact expected string.

> expect.stringMatching(string | regexp) matches the received value if it is a string that matches the expected string or regular expression.

Also, I'm a little wigged out there there is both documentation at https://jest-bot.github.io/jest/ and https://jestjs.io. I assume that jestjs.io is more up to date and official?

kalvinwang: By "a string that contains the exact expected string" it doesn't mean exact matches, it means partial matches. (I think the docs should drop the word "exact" because it only adds confusion/ambiguity)

I tested this by replacing:
```
      const url = getUrl(linkKey)
      expect(url).toEqual(expect.stringMatching(mockedPrefix))
```
with:
```
      const url = 'mock-prefix-blah'
      expect(url).toEqual(expect.stringContaining(mockedPrefix))
```
and the test still passes when running `yarn test tests/utils/getUrl.test.ts` which is what you'd expect if stringContaining is for partial matches

(Oh yeah, jestjs.io looks more official! I was top-result googling 😅)

rocketnova: I am saddened by the language in the jest docs 😞

rocketnova: Resolved in #568

PR #563, opened 2021-11-29
Removes the unused Loading logic

lomky:
===

Resolves #231 

- [x] Loading functionality is removed from code, stories, and tests
- [x] Storybook still loads
- [x] App is visually & functionally unaffected outside of loading

PR #564, opened 2021-11-30
Tweaking Storybook layout & organization

lomky:
I started by just moving the `maintenance` story in #563 but got a bit out of hand.

 - Adds the Main storybook section under Page Section to sit beside Header & Footer
 - Moves the maintenance component to the Main area, next to Claim Section
 - Moves Claim Details & Claim Status to under Claim Section
 - defaults to centering components on the canvas

a lot of this is pure taste, so please push back if this is too much.
 
Before | After
--|--
flat story layout | sectioned story layout

- [x] storybook still works
- [x] storybook sidebar makes sense


.storybook/preview.js
```diff
@@ -34,7 +35,7 @@ export const parameters = {
         'Claim Status Tracker',
         ['Page', '404'],
         'Component',
-        ['Page Section', ['Header', 'Main', 'Claim Section', 'Claim Status', 'Next Steps', 'Claim Details', 'Footer']],
```
kalvinwang, line 37: I assume this is intentional to remove Next Steps, but why doesn't it show up in the "before" screenshot?

kalvinwang (approved):
lgtm! style-wise you can wait for Rocket's opinion since I haven't used storybook enough to have strong opinions, but it doesn't seem too drastic to me :p 

(I think in general I'd prefer less hiding, so instead of nesting Claim Status and Details inside another Claim Section, I'd just do "Claim Section - Status" and "Claim Section - Details" at the parent level... but it's 100% up to you since you've been using it!)

.storybook/preview.js
lomky: it was a remnant from awhile ago - it didn't do anything

rocketnova (approved):
I don't feel strongly about this. I can confirm that storybook still works and that the sidebar makes sense. I like the nesting. I think it really helps someone who is less intimately familiar with the page structure understand how the different components relate to one another.

My only nit is that we have a story called `Claim Section` and then a folder right underneath it also called `Claim Section`. I think the duplication is confusing.
Screen Shot 2021-11-30 at 11 51 24 AM

PR #572, opened 2021-12-03
Migrate CSS to EDD template. Incorporate EDD template JS.

### LINKED ISSUE #118: Refine the Footer Component
### Description

As implemented in #74 the behavior of the footer doesn't quite match [EDD](https://edd.ca.gov/), showing a single column instead of two. This ticket is to more perfectly match the experience of the main EDD site.

### Acceptance Criteria

  - [ ] Our footer matches the EDD footer on smaller screens

rocketnova:
## Ticket

Resolves #118 

## Changes

Refactors CST front-end to align with EDD's v1.0.0 webstandards template:

- Refactors CST components to leverage the template
- Refactors custom CST css to use the template instead where possible
- Adds CA gov core JS 
- Adds jquery to support CA gov core JS
- Reverts oceanside theme to v5.0.15 to align with EDD's web standards template

## Context

We were notified last week that EDD released v1.0.0 of their webstandards template. This PR brings CST in line with those standards. This fixes the issues we were seeing with the footer in #118. 

This PR also makes it easier for long-term support of CST by making updates by making the front-end context much more similar to other EDD sites. Updates to the CSS will be (hopefully) as easy as just replacing the upstream CSS and JS files with new released versions.

## Testing

- [x] Review [chromatic/storybook build](https://60705d04dcad7600211e34d2-ijpmlixoib.chromatic.com) to ensure all elements of CST still render as expected
- [x] Compare behavior against [an example EDD page](https://www.edd.ca.gov/About_EDD/Accessibility.htm) to ensure the same behavior occurs for major elements, such as:
  - [x] Global header at page load (load at each breakpoint, open and close on mobile)
  - [x] Global header when scrolling (there is a sticky header feature in this design)
  - [x] Footer
  - [x] Header/footer links, hover, focus
  - [x] Each breakpoint and mobile, including window resize
- [x] Make sure this error doesn't show up when running `yarn dev` or `yarn start`
Screen Shot 2021-12-02 at 2 32 05 PM


.storybook/preview-head.html
```diff
@@ -4,3 +4,16 @@
   href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:ital,wght@0,400;0,700;1,400;1,700&display=swap"
   rel="stylesheet"
 />
+<!-- Import EDD template js -->
+<script type="text/javascript" src="/js/libs/jquery.js"></script>
+<script type="text/javascript" src="/js/cagov.core.js"></script>
```
rocketnova, line 9: @lomky Please let me know if you can think of any issues with including these in storybook.

.storybook/preview-head.html
```diff
@@ -4,3 +4,16 @@
...
+
+<!-- Override default storybook style to make it compatible with EDD template -->
+<style>
+  .sb-show-main.sb-main-centered {
+    display: block !important;
+  }
+  .sb-show-main.sb-main-centered #root {
+    padding: 0 !important;
+  }
+</style>
```
rocketnova, line 19: @lomky These are the css overrides I made to make the rendering work better. These are applied to the storybook wrapper divs. Let me know if you think there are any issues with these.

.storybook/preview-head.html
lomky: nothing specific jumps to mind

.storybook/preview-head.html
lomky: my only concern is if we think this could disguise changes to the application CSS, and I think its very clearly scoped to storybook-only. so good by me

lomky (changes requested):
This looks so great! Only thing to bring back is the vertical spacing within the `main` section of the Claim Section - those were intentional whitespacing choices from design around info density, which I think we want to preserve even above the EDD norms.

lomky (approved):
Nice work! the behavior is so much smoother

PR #581, opened 2021-12-13
Update help link to be different for English and Spanish

kalvinwang:
Updates help link to be different for English and Spanish

From: https://uio.edd.ca.gov/UIO/Pages/Public/help/index.htm#t=en-US/Public/NewClaim/UIOnlineNewClaimLandingPage.htm 
To: https://uio.edd.ca.gov/UIO/Pages/Public/help/index.htm#t=es-MX/Public/NewClaim/UIOnlineNewClaimLandingPage.htm

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -95,7 +95,7 @@ export const Header: React.FC<HeaderProps> = ({ userArrivedFromUioMobile = false
                 <span className="text">{t('header.edd-home')}</span>
               </Nav.Link>
             </Navbar.Collapse>
-            <Nav.Link target="_blank" rel="noopener noreferrer" href={getUrl('uio-desktop-help-new-claim')}>
+            <Nav.Link target="_blank" rel="noopener noreferrer" href={t('urls.header.uio-desktop-help-new-claim')}>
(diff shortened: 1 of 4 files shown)


rocketnova (changes requested):
@kalvinwang This looks wrong to me. This is a uio url, which would presumably need to be different for each lower environment based on the `URL_PREFIX_*` env vars. That means that the url should come from `urls.json`, not the language-specific urls. 

If the only difference between thes two urls is the `en-US` and `es-MX` portion, then I would suggest doing a string replacement in `getUrl.ts`. In `Header.tsx`, we get the language from `i18n`, see line 37. 

The call to the `getUrl()` function would become something like `getUrl('uio-desktop-help-new-claim', urlPrefixes, i18n.language)`. Then we would do a replacement in `getUrl()` like this:

```typescript
if (lang == 'es') {
  const pattern = /t=.{2}-.{2}\//
  urls[key] = urls[key].replace(pattern, 't=es-MX/')
}
```

kalvinwang:
Thanks Rocket! Started a new PR, #582

